PR #4, opened 2014-04-20
MessagesStorage & RequestStorage

MartinMajor:
This pull request separates flash messages into new service MessagesStorage. It also separates store & restore requests into new service RequestStorage. This new service redirects to original URL when restoring old requests.


f3l1x:
:+1:

PR #16, opened 2014-06-23
Added presenter mapper

Budry:
Discusion: http://forum.nette.org/en/19620-better-mapping-configuration#p135799

For easier use custom mapping implementation

_with https://github.com/nette/bootstrap/pull/8_

### DIFF
--- a/src/Application/IPresenterMapper.php
+++ b/src/Application/IPresenterMapper.php
@@ -1,0 +1,25 @@
+<?php
+
+/**
+ * This file is part of the Nette Framework (http://nette.org)
+ * Copyright (c) 2004 David Grudl (http://davidgrudl.com)
+ */
+
+namespace Nette\Application;
+
+interface IPresenterMapper
+{
+	/**
+	 * Formats presenter class name from its name.
+	 * @param  string
+	 * @return string
+	 */
+	function formatPresenterClass($presenter);
+
+	/**
+	 * Formats presenter name from class name.
+	 * @param  string
+	 * @return string
+	 */
+	function unformatPresenterClass($class);
+}
--- a/src/Application/PresenterFactory.php
+++ b/src/Application/PresenterFactory.php
@@ -20,22 +20,20 @@ class PresenterFactory extends Nette\Object implements IPresenterFactory
 	/** @var bool */
 	public $caseSensitive = FALSE;
 
-	/** @var array[] of module => splited mask */
-	private $mapping = array(
-		'*' => array('', '*Module\\', '*Presenter'),
-		'Nette' => array('NetteModule\\', '*\\', '*Presenter'),
-	);
-
 	/** @var array */
 	private $cache = array();
 
 	/** @var Nette\DI\Container */
 	private $container;
 
+	/** @var \Nette\Application\IPresenterMapper */
+	private $presenterMapper;
 
-	public function __construct(Nette\DI\Container $container)
+
+	public function __construct(Nette\DI\Container $container, IPresenterMapper $presenterMapper)
 	{
 		$this->container = $container;
+		$this->presenterMapper = $presenterMapper;
 	}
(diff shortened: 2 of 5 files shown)


dg:
ping @Budry

Budry:
I apologize for the delay, but I lived in the belief that this pull is dead. 

Problem is that actual mapping implementation isn't really usable for specific cases. For example (from discussion) with actual mapping implementation I don't use namespaces as `App\BlogModule\Account\Presenters\FrontPresenter` i'm not sure if it is possible (I think that it is not possible). It would be good to be able to use a placeholder for each module ([to see](http://forum.nette.org/en/19620-better-mapping-configuration#p135799)).

Currently I'm not sure if it necessary to change current implementation, but able to use a custom mapper can be very good and it can solved problem with specific requirements for mapping.

I can update this pull if you say that it makes sense

dg:
To add Mapper is to add two new classes, which will be always autoloaded. Some people scream when you add one class [that is always autoloaded](https://github.com/nette/safe-stream/commit/313134b8d511469291dd946d5416485b90c159a3#commitcomment-9380744) (even it is autoloaded only if you install this package).

To add two classes increases the complexity of the framework. I don't want to increase the complexity if it solves use case for only one person. But I want to find any solution, same case [as here](https://github.com/nette/application/pull/19).

So for your use case is solution to create own IPresenterFactory or descendant of default PresenterFactory.

xificurk:
> I don't want to increase complexity if it solves use case for only one person.

Actually, I would welcome this as well. It would come really handy for overriding generic presenters shipped with installed package by app-specific presenter (see https://github.com/nette/application/pull/56#issuecomment-71458691).
Now the choice for us is either re-implement whole `PresenterFactory` with our own mapping solution, or inherit the `PresenterFactory` from Nette with unwanted/unused default mapping and hack over it.

dg:
It this case it would be better to prepare an RFC and open discussion about it and we can implement it in 2.4.

Budry:
I'm for open discussion :+1:

> I don't want to increase complexity if it solves use case for only one person

I understand and I'm not saying that it is necessary implement this, but this is not only my case (for example http://forum.nette.org/cs/17194-jak-nastavit-presenter-mapping-pro-class-name-app-blablamodule-blabla-blablapresenter and were more, but now I don't find it). 

This pull is very old and was created from poor application design, but still I think that current mapping implementation is too restrictive.

dg:
What exactly you mean by "too restrictive"? Yes, it is limited, but extracting it to new class is not solution - the new DefaultMapper wil be exactly the same restrictive as current PresenterFactory (https://github.com/nette/application/pull/16#issuecomment-48160051).

Budry:
You are right. Separated mapper into a new class is not solutions. I think that would be better change current mapping implementation and make it less limited. Be able to use a construction similar to this 
```
App\<module>Section\<module>\Presenters\<presenter>Presenter
:Blog:Account:Admin:default => App\BlogSection\Account\Presenter\AdminPresenter
```
would be very good. 

I created this pull because I know a lot of people and project which implement custom `PresenterFactory` because they need mapping similar to example and I didn't think that implement custom `PresenterFactory` is  good way but change only Mapper yes (it is better way in my opinion), but as I said the best solution would be change mapping implementation and if someone needs more specific way may use their own `PresenterFactory`

I haven't problem to close this pull and open discussion about better mapping implementation :)

dg:
To change title is probably sufficient, here is history, or not?

Budry:
I was not sure if this is pull a suitable place for discussion about new mapping and I wanted to establish a topic on the forum

dg:
https://forum.nette.org/cs/26116-presentermapper-ipresenterformatter-apod

TomasVotruba:
Seems legit. What needs to be done here?

PR #19, opened 2014-07-06
Route: Fixed handling of object parameters

enumag:
Sometimes I use objects as parameters. A nice way to deal with them is to put them into another object using this class: https://gist.github.com/enumag/1f98ce3beffa2490c6ca.

Using this the object will be converted to it's identifier using the __toString method unless I do something else with it in the FILTER_OUT function. First way is useful in AdminModule because I don't care about pretty urls there. Second way is handy for FrontModule because I can slugify the object in FILTER_OUT.

This works fine if the parameter is mentioned in the route string directly. If it isn't it should appear in the query string - and this is the problem. The parameter disappears instead because the `http_build_query` function ignores objects completely (even if they have a __toString method declared).

### DIFF
--- a/src/Application/Routers/IObjectParameter.php
+++ b/src/Application/Routers/IObjectParameter.php
@@ -1,0 +1,19 @@
+<?php
+
+/**
+ * This file is part of the Nette Framework (http://nette.org)
+ * Copyright (c) 2004 David Grudl (http://davidgrudl.com)
+ */
+
+namespace Nette\Application\Routers;
+
+
+interface IObjectParameter
+{
+
+	/**
+	 * @return string representation of the parameter
+	 */
+	function __toString();
+
+}
--- a/src/Application/Routers/Route.php
+++ b/src/Application/Routers/Route.php
@@ -338,6 +338,12 @@ class Route extends Nette\Object implements Application\IRouter
 			}
 		}
 
+		array_walk_recursive($params, function (&$value) {
+			if ($value instanceof IObjectParameter) {
+				$value = (string) $value;
+			}
+		});
+
 		// compositing path
 		$sequence = $this->sequence;
 		$brackets = array();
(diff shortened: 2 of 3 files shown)


dg:
@enumag this can be handled via global filters, or not? http://doc.nette.org/en/2.2/routing#toc-filters-and-translation

enumag:
@dg Unfortunately not. My use-case is that I need the original object in the parameter-specific FILTER_OUT functions - to replace the object by it's slug instead of id. Global FILTER_OUT would be the solution only if it were called after the parameter-specific filters which is not the case.

enumag:
@dg Maybe I could use the global FILTER_OUT if I somehow knew which parameters are in mask and which are not. The ones in mask don't need to be replaced because __toString works for them and I don't need FILTER_OUT functions for those that are not in mask.

dg:
So can I close it?

enumag:
I don't have any way to know which parameters are in the mask, do I?

dg:
Via API? No. Maybe via `strpos($route->getMask(), "<$param")`

enumag:
@dg I thought about that but it doesn't look like something that would always work without false detections. Besides using the global FILTER_OUT is a work-around that might work for me now, but it is not a real solution.

dg:
Why not? It looks like ideal solution.

enumag:
Doesn't feel like it to me. Anyway `strpos` is not the way because some parameter name might be prefix of another so I'll have to try some regular expressions or maybe I'll send a PR with new method that would get the info from route metadata. Let me try it out and I'll close this issue if it works.

fprochazka:
If the `setMask` analyzer was split to several protected methods, it should be easier to override one of them and because it already parses the mask, it would be a reliable way of telling if the parameter is really there.

enumag:
@fprochazka For my use-case it's actually even better to not look at which parameters are in the mask but which parameters have custom filter out funciton. See #62 for implementation.

PR #20, opened 2014-07-22
Enhanced searching for templates WIP

dg:
See test https://github.com/nette/application/pull/20/files#diff-225c556e420e89cd621ebb39127fe77f


TomasVotruba:
Is this still relevant?
I think it can be closed maybe.

dg:
I think it is cool :)

TomasVotruba:
I don't understand what it does :(

> I think it is cool :)

So why not merge it?

f3l1x:
@dg It looks for layout and template of parent presenter? Could you please give us an example? :-)

mabar:
We have something similar in contributte/application
https://github.com/contributte/application/blob/master/src/UI/Presenter/StructuredTemplates.php

PR #35, opened 2014-10-05
Router: optimized for a single immutable $refUrl instance

JanTvrdik:
This makes it about 40 % faster with the following benchmark. However it relies on `$refUrl` being immutable.

```php
$appRequest = new AppRequest('Admin', 'GET', ['action' => 'default']);
$refUrl = new UrlScript('http://localhost/');

$time = -microtime(true);
$routerB = new RouteList();
$routerB[] = new Route('', 'Homepage:default');
$routerB[] = new Route('sign-in', 'Auth:signIn');
$routerB[] = new Route('sign-out', 'Auth:signOut');
$routerB[] = new Route('sign-up', 'Auth:signUp');
$routerB[] = new Route('sign-up-confirm', 'Auth:signUpConfirm');
$routerB[] = new Route('forgotten-pass', 'Auth:forgottenPassword');
$routerB[] = new Route('view', 'Product:view');
$routerB[] = new Route('manage', 'Product:manage');
$routerB[] = new Route('dashboard', 'Dashboard:default');
$routerB[] = new Route('admin', 'Admin:default');

for ($i = 0; $i < 10000; $i++) {
	$routerB->constructUrl($appRequest, $refUrl);
	// $routerB->match($httpRequest);
}
$time += microtime(true);

echo round($time * 1000, 2) . " ms\n";
```

### DIFF
--- a/src/Application/Routers/Route.php
+++ b/src/Application/Routers/Route.php
@@ -108,6 +108,12 @@ class Route extends Nette\Object implements Application\IRouter
 	/** @var int */
 	private $flags;
 
+	/** @var Nette\Http\Url */
+	private $lastRefUrl;
+
+	/** @var string */
+	private $lastBaseUrl;
+
 
 	/**
 	 * @param  string  URL mask, e.g. '<presenter>/<action>/<id \d{1,3}>'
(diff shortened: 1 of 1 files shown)


dg:
:+1: but we should add to the contract that `Nette\Http\Url` must be always the same.

JanTvrdik:
@dg Why? It works well with multiple `Nette\Http\Url` instances. It does not cause any BC break. It was always assumed the `Nette\Http\Url` is immutable, e.g. presenter [relies on this immutability](https://github.com/nette/application/blob/2d1e6b9880a3b375ce65df57ca3715850b6c0ecc/src/Application/UI/Presenter.php#L977-L980) for a very long time.

dg:
ok

PR #40, opened 2014-11-02
Generalized Route::getTargetPresenter() 

JanTvrdik:
Allow using this optimization for other `IRouter` implementations.

Ideas for better names of anything are welcomed.

### DIFF
--- a/src/Application/Routers/RouteList.php
+++ b/src/Application/Routers/RouteList.php
@@ -96,8 +96,9 @@ class RouteList extends Nette\Utils\ArrayList implements Nette\Application\IRout
 	{
 		if (!$route instanceof Nette\Application\IRouter) {
 			throw new Nette\InvalidArgumentException('Argument must be IRouter descendant.');
+		} elseif ($this->cachedRoutes !== NULL) {
+			throw new \LogicException('No!'); // TODO: we need some LogicException from Nette namespace
 		}
-		$this->cachedRoutes = NULL;
 		parent::offsetSet($index, $route);
 	}
 


dg:
I think that `getTargetPresenters` can be part of IRouter.

JanTvrdik:
That would solve the ugly `IFixedTargetPresenters` name. But it would be a nasty BC break. It would actually be so large BC break that if we choose to do it, we may as well break other things (I have few ideas).

dg:
I would prefer to avoid BC break. This method can be in interface in v2.3 commented out and checked via method_exists.

JanTvrdik:
That seems reasonable.

dg:
Typo in commit message: IRouter

JanTvrdik:
Yeah and it also does not have tests so we don't know whether it actually work.

JanTvrdik:
Do you plan to merge this to 2.3?

dg:
Yes, will you finish it?

dg:
ping @JanTvrdik

JanTvrdik:
There is a potential issue with cache invalidation when combining multiple `RouteList`s, i.e. the following test will fail:

~~~php
$listA = new RouteList();
	$listB = new RouteList();

	Assert::same([], $listA->getTargetPresenters());
	Assert::same([], $listB->getTargetPresenters());

	$listB[] = new Route('about', 'About:default');

	Assert::same(['About'], $listA->getTargetPresenters());
	Assert::same(['About'], $listB->getTargetPresenters());
~~~

JanTvrdik:
One possible solution may be to disallow (throw exception) modifying `RouteList` once first URL has been generated (or more precisely once the cache is built).

chemix:
ping @JanTvrdik would you like to merge to 2.4 version?

JanTvrdik:
Yes. It would be awesome if anyone could finish this without me (I don't currently have the mental model for this ready anyway) but I can finish this if no one else finds the time.

PR #43, opened 2014-12-15
new routers: DomainRouter and ModuleRouter

JanTvrdik:
Refactoring based on https://github.com/nextras/static-router/issues/2

Obviously, I will write tests and possibly handle the BC break better if the idea is accepted.

### DIFF
--- a/src/Application/Routers/DomainRouter.php
+++ b/src/Application/Routers/DomainRouter.php
@@ -30,12 +30,12 @@ class DomainRouter extends Nette\Object implements Nette\Application\IRouter
 
 	/**
 	 * @param  string
-	 * @param  Nette\Application\IRouter
+	 * @param  Nette\Application\IRouter|Nette\Application\IRouter[]
 	 */
-	public function __construct($host, Nette\Application\IRouter $innerRouter)
+	public function __construct($host, $innerRouter)
 	{
 		$this->host = $host;
-		$this->innerRouter = $innerRouter;
+		$this->innerRouter = is_array($innerRouter) ? new RouteList($innerRouter) : $innerRouter;
 	}
 
 
--- a/src/Application/Routers/DomainRouter.php
+++ b/src/Application/Routers/DomainRouter.php
@@ -1,0 +1,78 @@
+<?php
+
+/**
+ * This file is part of the Nette Framework (http://nette.org)
+ * Copyright (c) 2004 David Grudl (http://davidgrudl.com)
+ */
+
+namespace Nette\Application\Routers;
+
+use Nette;
+
+
+/**
+ * @property-read string $host
+ */
+class DomainRouter extends Nette\Object implements Nette\Application\IRouter
+{
+	/** @var string */
+	private $host;
+
+	/** @var Nette\Application\IRouter */
+	private $innerRouter;
+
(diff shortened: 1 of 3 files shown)


JanTvrdik:
Any thoughts, @dg? It's been a month.

dg:
I haven't thought about it yet.

foxycode:
I like both new routers. I actually using routes for more subdomains and it is annoying write `something.%domain%` everywhere. :+1:

EdaCZ:
Why is this closed, please?

This PR seems like a good idea to me.

dg:
Github fools you ;)

EdaCZ:
Oh. Now I see... Great :-)

JanTvrdik:
ping @dg What's the status?

fprochazka:
This is not sufficient. Your router is way too strict. I'm preparing a modified version of your router that works better. Just needs s bit of polishing.

fprochazka:
I've ended up using this version on production. https://gist.github.com/fprochazka/769596173ee4e1fc02b2

- the first argument is expression that will be used if the domain of http request doesn't match the other arguments, so the router is still able to generate url
- the second argument is used for allowing the current httprequest domain to be used for constructing urls. For example you have `testing.` subdomain or `beta.` and the app should be able to work with that without redirecting to `www.`
- the third argument is for matching domains

f3l1x:
:+1: for DomainRouter and ModuleRouter

JanTvrdik:
self note: collision between URL and presenter tree.

mabar:
Is there any chance it will be merged in 3.0? I was creating something similar recently and I absolutely like this idea.

f3l1x:
I like an idea too. I think I gonna create some classes in [contributte/application](https://github.com/contributte/application).

dg:
I want to add to RouteList syntactic sugar similar to forms, ie. instead of

```php
$router = new RouteList();
$router[] = new Route(...);
$router[] = new Route(...);
$router[] = new MyRouter(...);
```

to allow something like

```php
$router = new RouteList();
$router->addRoute(...);
$router->addRoute(...);
$router->add(new MyRouter...);
```

Because there is no need to deal with class names so much. And classes can be confusing after merging #208.

In relation to this PR original code:

```php
$router = new RouteList();
$router[] = new DomainRouter('blog.example.com', new RouteList([
    new ModuleRouter('Admin', new RouteList([
        new Route('...'),
        new Route('...'),
    ])),
    new ModuleRouter('Front', new RouteList([
        new Route('...'),
        new Route('...'),
    ])),
]));

$router[] = new DomainRouter(
    'simple.example.com',
    new SimpleRouter(),
);
```

can be written this way:

```php
$router = new RouteList();
$router->withDomain('blog.example.com')
	withModule('Admin')
		->addRoute('...')
		->addRoute('...')
	end() // or store $blogRouter = $router->withDomain('blog.example.com'); and continue with $blogRouter->withModule(...)
	withModule('Front')
		->addRoute('...')
		->addRoute('...');

$router->withDomain('simple.example.com')
	->add(new SimpleRouter());
```

What do you think?

f3l1x:
I like it! 👍

dg:
Closed by #208

PR #47, opened 2014-12-17
LatteMacros: snippet rendering support for in template defined variables

### LINKED ISSUE #46: Support for global parameters in latte template in snippets
I tried to define „global variable“ in one snippet for usage in another snippet ([example](http://forum.nette.org/cs/21594-proof-of-concept-v-sablone-nastavitelnych-sablon-anew-latte-makro-template#p147845)) but there is a little block in [Application/Bridges/ApplicationLatte/UiMacros.php](https://github.com/nette/application/blob/master/src/Bridges/ApplicationLatte/UIMacros.php#L53).

There is calling 
```php
return Nette\Bridges\ApplicationLatte\UIMacros::renderSnippets($_control, $_b, get_defined_vars());
```

But if there will be 
```php
return Nette\Bridges\ApplicationLatte\UIMacros::renderSnippets($_control, $_b, $template);
```

and in the implementation of renderSnippets() method will be $template->getParameters() instead of [current usage](https://github.com/nette/application/blob/master/src/Bridges/ApplicationLatte/UIMacros.php#L122), it will be possible to define „global“ variable in one snippet by `{?$template->variable = 'value'}` or by new simple latte macro `{varGlobal $variable = 'value'}` and it will be accessible like when it will be defined before render.

Can I prepare pull-request?

zeleznypa:
Currently you can access `$template->getParameters()` defined in presenter/control in the snippet directly.

For example, if the `$template->getParameters()` contain `array('test' => 'value')`, you can access this variable in snippet directly:
```latte 
{snippet test}
    {dump $test}
{/snippet}
```

`$template` available in the latte is an object of `Latte\Template`, that is used to render current template. So when you set additional variable in this object, you can access this variable elsewhere in the same template and also in the snippets.

So when you have following code, everything work well.
```latte
{var $template->another_test = 'another value'}
{snippet test}
    {dump $template->another_test}
{/snippet}
```

But all parameters from `$template->getParameters()` can be accessible directly,
```latte
{var $template->another_test = 'another value'}
{snippet test}
    {dump $another_test}
{/snippet}
```
but except the in template defined.

This pull-request allow to define "in template global variable", than can be accessible directly in snippets like in the last example.

Also it resolve #46

### DIFF
--- a/src/Bridges/ApplicationLatte/UIMacros.php
+++ b/src/Bridges/ApplicationLatte/UIMacros.php
@@ -50,7 +50,7 @@ class UIMacros extends Latte\Macros\MacroSet
 		$prolog = '
 // snippets support
 if (empty($_l->extends) && !empty($_control->snippetMode)) {
-	return Nette\Bridges\ApplicationLatte\UIMacros::renderSnippets($_control, $_b, get_defined_vars());
+	return Nette\Bridges\ApplicationLatte\UIMacros::renderSnippets($_control, $_b, $template);
 }';
 		return array($prolog, '');
 	}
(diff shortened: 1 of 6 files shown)


dg:
You should use `{snippetArea}`

PR #49, opened 2015-01-06
Presenter: fixed usage of immutable Url

stekycz:
It can be changed whenever so this is future compatibility.

Related to https://github.com/nette/http/pull/43


JanTvrdik:
:-1: This is opened unnecessary soon.

PR #50, opened 2015-01-08
JsonResponse::__construct() gets charset as optional parameter

mkoubik:


Majkl578:
:-1:, as per RFC 4627, JSON SHALL (must) be encoded in Unicode.

mkoubik:
Cool, I didn't know.

JanTvrdik:
@Majkl578 Unicode is not an encoding. JSON may be encoded in UTF-16.

So I'm :+1: on this RFC.

src/Application/Responses/JsonResponse.php
```diff
@@ -26,18 +26,23 @@ class JsonResponse extends Nette\Object implements Nette\Application\IResponse
...
+
 
 	/**
 	 * @param  array|\stdClass  payload
-	 * @param  string    MIME content type
+	 * @param  string           MIME content type
+	 * @param  string           charset
 	 */
-	public function __construct($payload, $contentType = NULL)
+	public function __construct($payload, $contentType = NULL, $charset = NULL)
```
JanTvrdik: What about making UTF-8 a default `$charset` value?

mkoubik:
@JanTvrdik:
```
Since the first two characters of a JSON text will always be ASCII
   characters [RFC0020], it is possible to determine whether an octet
   stream is UTF-8, UTF-16 (BE or LE), or UTF-32 (BE or LE) by looking
   at the pattern of nulls in the first four octets.
```

src/Application/Responses/JsonResponse.php
JanTvrdik: Oh, I did not realize the due to using `json_encode` the encoding will always be UTF-8.

Majkl578:
@JanTvrdik: I know, but I've never seen other than UTF-8 in real life so far, at least in PHP.

PR #52, opened 2015-01-18
added LinkGenerator

dg:
Similar to https://github.com/nextras/link-factory. It should be part of framework. Solves an issue with generating links in emails.

### DIFF
--- a/src/Application/LinkGenerator.php
+++ b/src/Application/LinkGenerator.php
@@ -1,0 +1,79 @@
+<?php
+
+/**
+ * This file is part of the Nette Framework (http://nette.org)
+ * Copyright (c) 2004 David Grudl (http://davidgrudl.com)
+ */
+
+namespace Nette\Application;
+
+use Nette;
+
+
+/**
+ * Link generator.
+ */
+class LinkGenerator extends Nette\Object
+{
+	/** @var IRouter */
+	private $router;
+
+	/** @var Nette\Http\Url */
+	private $refUrl;
+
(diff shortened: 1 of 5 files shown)


JanTvrdik:
* Another name to consider – `Linker`.
* The code is hard to read for no good reason due to aggressive inlining.
* I don't like the dependency on `Application\UI`. We may even consider putting this to `nette/routing` if you ever choose to [split nette/application](http://forum.nette.org/en/21178-rfc-refactoring-nette-application-dependencies).

fprochazka:
Adding it to routing seems like a good idea.

dg:
"Linker" is something similar to compiler.

dg:
Split nette/application to nette/routing will take some time, so it is not goal for 2.3.

JanTvrdik:
@dg I know but it should be taken into consideration.

JanTvrdik:
Because you decided to write it more generally (support presenters without `action`) parameter, it may be nice to allow any value in `action` parameter, ie.

~~~php
if (!preg_match(...)) {
    throw new UI\InvalidLinkException(...);
} elseif ($m[2]) {
    $params[UI\Presenter::ACTION_KEY] = $m[2]
}
~~~

This would allow programmers to write
~~~php
$generator->link('Foo:', ['action' => $arr]);
~~~

dg:
Good point

Caspern:
Wouldn't be better to generate relative links by default and allow ```//Presenter:default``` syntax for absolute ones?

PR #53, opened 2015-01-19
Extensions: used services and config aliases [WIP]

dg:
See nette/di#29 and nette/bootstrap#32

### DIFF
--- a/src/Bridges/ApplicationDI/ApplicationExtension.php
+++ b/src/Bridges/ApplicationDI/ApplicationExtension.php
@@ -33,19 +33,10 @@ class ApplicationExtension extends Nette\DI\CompilerExtension
 
 	public function loadConfiguration()
 	{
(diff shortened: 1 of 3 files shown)


dg:
I have absolutely no idea how to define, that a bridge requires nette/di in some version…

xificurk:
I think current requirement of `"nette/di": "~2.3"` is just fine. What's the problem?

dg:
nette/application requires DI, so it is fine. But I have problem with nette/database.

xificurk:
I'm for adding conflict to composer.json, see proposed nette/mail#6

dg:
It is good idea. And why conflict with whole nette/nette 2.2?

xificurk:
It's "<2.2"... I haven't changed that ;)

dg:
Mea culpa ;)

PR #55, opened 2015-01-25
Remove nette/reflection

dg:


JanTvrdik:
Good job! Another thing I can cross off my todolist.

dg:
I think revert actually fixes this bug, see original commit https://github.com/nette/nette/commit/35bb0ce4beaa9ba9f71e28ff3a09d8c236aafce6

dg:
I changed this PR, PresenterComponentReflection still extends `Nette\Reflection\ClassType` to maintain BC.

JanTvrdik:
@dg `PresenterComponentReflection` is `@internal` class, what exactly would be broken?

dg:
I think that nothing, but just for sure I'll do this change in 2.4-dev.

JanTvrdik:
That seems like a strange reason to keep the dependency. How is the situation here different from nette/di where you have removed the dependency?

enumag:
@JanTvrdik The difference is `PresenterComponent::getReflection()`. Even now `Object::getReflection()` still returns `Nette\Reflection\ClassType` which would break for PresenterComponent.

PR #57, opened 2015-01-29
MicroPresenter - error during display text

f3l1x:
This example: 

```php
$router[] = new Route('<path .*>', function ($presenter, $path) {
    return '{=time()}';
});
```

Triggers 2 errors.

* Fatal error
* Logical error

See more in commit message.


JanTvrdik:
Few points:

* missing unit test
* variable names should not start with underscore. Something like `$templateSource` and `$templateParams` could be used.

src/Application/MicroPresenter.php
```diff
@@ -93,11 +94,13 @@ public function run(Application\Request $request)
...
-			if ($response[0] instanceof \SplFileInfo) {
-				$response->setFile($response[0]);
+			list($_content, $_parameters) = $response;
+			$response = $this->createTemplate()->setParameters($_parameters);
+			if ($_content instanceof \SplFileInfo) {
+				$response->setFile($_content);
 			} else {
-				$response->setSource($response[0]);
+				$response->getLatte()->setLoader(new Latte\Loaders\StringLoader);
+				$response->setFile($_content);
```
JanTvrdik: This could be simplified:

~~~php
if (is_string($_content)) {
    $response->getLatte()->setLoader(new Latte\Loaders\StringLoader);
}
$response->setFile($_content);
~~~

dg:
And try to write better commit message please.

f3l1x:
I changed commit message and added test cases. Hope, it is better now.

enumag:
@f3l1x Newlines are missing at the end of new files.

f3l1x:
What about now?

dg:
Thank you, merged

f3l1x:
Thanks. :+1:

PR #59, opened 2015-02-01
Fixed routing debugger

enumag:
It's better for the future to not rely on the application alias. Also it's better to use `ContainerBuilder::getByType()` to find the service because the name of ApplicationExtension can vary as well. Of course `ContainerBuilder::getByType()` doesn't work in loadConfiguration, hence I moved the code to beforeCompile.


dg:
Please write consistent commit message…

dg:
Thx!

PR #60, opened 2015-02-02
Fixed RoutingExtension

enumag:
This is my mistake of course. I'm not used to using the validateConfig method yet so I just copied it without looking what it actually does.


dg:
squased https://github.com/nette/application/commit/60b11f5bf654d05f8d4527f16c670d96860d8a17, fixed commit message

enumag:
thx

PR #65, opened 2015-02-22
InvalidLinkMode: changed invalid link handling

klimesf:
[See the RFC](http://forum.nette.org/en/22153-link-and-plink-macros-produce-e-warning-if-link-is-not-valid)

The build isn't failing due to the changes I made. I can't address the issue in this PR.

### DIFF
--- a/src/Application/UI/Presenter.php
+++ b/src/Application/UI/Presenter.php
@@ -36,7 +36,10 @@ abstract class Presenter extends Control implements Application\IPresenter
 	/** bad link handling {@link Presenter::$invalidLinkMode} */
 	const INVALID_LINK_SILENT = 1,
 		INVALID_LINK_WARNING = 2,
-		INVALID_LINK_EXCEPTION = 3;
+		INVALID_LINK_VISUAL = 4;
+
+	/** @deprecated Please use INVALID_LINK_WARNING instead. */
+	const INVALID_LINK_EXCEPTION = 3;
 
 	/** @internal special parameter key */
 	const SIGNAL_KEY = 'do',
@@ -1055,19 +1058,23 @@ abstract class Presenter extends Control implements Application\IPresenter
 
 	/**
 	 * Invalid link handler. Descendant can override this method to change default behaviour.
+	 * @param InvalidLinkException $e
 	 * @return string
 	 * @throws InvalidLinkException
 	 */
 	protected function handleInvalidLink(InvalidLinkException $e)
 	{
-		if ($this->invalidLinkMode === self::INVALID_LINK_SILENT) {
-			return '#';
-
-		} elseif ($this->invalidLinkMode === self::INVALID_LINK_WARNING) {
-			return '#error: ' . $e->getMessage();
-
-		} else { // self::INVALID_LINK_EXCEPTION
-			throw $e;
+		switch ($this->invalidLinkMode) {
+			case self::INVALID_LINK_EXCEPTION:
+				throw $e;
+			case self::INVALID_LINK_WARNING:
+				trigger_error('Invalid link: ' . $e->getMessage(), E_USER_WARNING);
+			case self::INVALID_LINK_VISUAL:
+				return '#error: ' . $e->getMessage();
+			case self::INVALID_LINK_SILENT:
+			default:
+				trigger_error('Invalid link: ' . $e->getMessage(), E_USER_WARNING);
+				return '#';
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


tests/Application.DI/ApplicationExtension.scan.phpt
```diff
@@ -89,3 +89,139 @@ test(function() {
...
+	eval($code);
+
+	/** @var Nette\DI\Container $container */
+	$container = new Container7;
+	$tags = $container->findByTag('nette.presenter');
+	Assert::count( 1, array_keys($tags, 'Presenter1') );
+	$tmp = array_keys($tags, 'Presenter1');
+	$presenter = $container->getService($tmp[0]);
+	Assert::same( \Nette\Application\UI\Presenter::INVALID_LINK_SILENT, $presenter->invalidLinkMode );
+});
```
klimesf: I will refactor this, WIP

src/Application/UI/Presenter.php
```diff
@@ -1055,19 +1058,21 @@ public static function argsToParams($class, $method, & $args, $supplemental = ar
...
-		if ($this->invalidLinkMode === self::INVALID_LINK_SILENT) {
-			return '#';
-
-		} elseif ($this->invalidLinkMode === self::INVALID_LINK_WARNING) {
-			return '#error: ' . $e->getMessage();
-
-		} else { // self::INVALID_LINK_EXCEPTION
-			throw $e;
+		switch($this->invalidLinkMode) {
+			case self::INVALID_LINK_WARNING:
```
fprochazka: Exception should not be dropped, it should produce at least a warning. Untill it's removed few versions later.

src/Application/UI/Presenter.php
```diff
@@ -1055,19 +1058,21 @@ public static function argsToParams($class, $method, & $args, $supplemental = ar
...
+				trigger_error('Invalid link: ' . $e->getMessage(), E_USER_WARNING);
+			case self::INVALID_LINK_VISUAL:
+				return '#error: ' . $e->getMessage();
+			case self::INVALID_LINK_SILENT:
+			default:
+				trigger_error('Invalid link: ' . $e->getMessage(), E_USER_WARNING);
```
fprochazka: why is silent producing an error?

src/Bridges/ApplicationDI/ApplicationExtension.php
```diff
@@ -141,4 +145,22 @@ private function findPresenters()
 		return $presenters;
 	}
 
+
+	/**
+	 * Sets invalidLinkMode for presenters.
+	 * @param bool
+	 */
+	private function setSilentInvalidLinks($silent = TRUE)
```
fprochazka: I don't like that you've used a magic setter for this.

src/Application/UI/Presenter.php
klimesf: This is described in the RFC. SILENT is meant for production and the name refers to the text it returns. E_USER_WARNING is produced so it's visible in the logs.

src/Application/UI/Presenter.php
klimesf: It behaves exactly like the INVALID_LINK_SILENT, see the `default` keyword below.

fprochazka: I don't like that it's not explicit.

src/Application/UI/Presenter.php
fprochazka: I see now :)

src/Bridges/ApplicationDI/ApplicationExtension.php
```diff
@@ -141,4 +145,22 @@ private function findPresenters()
 		return $presenters;
 	}
 
+
+	/**
+	 * Sets invalidLinkMode for presenters.
+	 * @param bool
+	 */
+	private function configureInvalidLinkMode($silent = true)
```
klimesf: @fprochazka Is it better this way? :)

klimesf:
If Presenter::invalidLinkMode is set to something else than the provided constants, it will now behave the same as INVALID_LINK_SILENT, whereas before it would behave as INVALID_LINK_EXCEPTION. Do you think this is a BC break?

src/Bridges/ApplicationDI/ApplicationExtension.php
kubinpetr: Use TRUE instead of true.

klimesf: Thank you! I missed this one.

dg:
What about something like this #66?

PR #66, opened 2015-02-25
PresenterFactory as callback & changed invalid link handling [WIP]

dg:
Two different things.

1) replaced PresenterFactory dependency on DI with simple callback. 
- dependency on DIC is bad (it is the only one dependency on DIC in whole framework)
- current PresenterFactory constructor is different in 2.2 and 2.3, so there will still be BC break
- partially solves #16 

2) new invalidLinkMode
- handling of invalid links is not insufficient, we need warnings
- I am not happy with #65, where (for partial combatibility) is mode that triggers warnings named `INVALID_LINK_SILENT`

### DIFF
--- a/src/Application/PresenterFactory.php
+++ b/src/Application/PresenterFactory.php
@@ -29,17 +29,16 @@ class PresenterFactory extends Nette\Object implements IPresenterFactory
 	/** @var array */
 	private $cache = array();
 
-	/** @var Nette\DI\Container */
-	private $container;
-
-	/** @var string */
-	private $touchToRebuild;
+	/** @var callable */
+	private $factory;
 
 
-	public function __construct(Nette\DI\Container $container, $touchToRebuild = NULL)
+	/**
+	 * @param  callable  function(string $class): IPresenter
+	 */
+	public function __construct($factory = NULL)
 	{
-		$this->container = $container;
-		$this->touchToRebuild = $touchToRebuild;
+		$this->factory = $factory ?: function($class) { return new $class; };
 	}
 
 
@@ -50,25 +49,7 @@ class PresenterFactory extends Nette\Object implements IPresenterFactory
 	 */
 	public function createPresenter($name)
 	{
-		$class = $this->getPresenterClass($name);
-		$services = array_keys($this->container->findByTag('nette.presenter'), $class);
-		if (count($services) > 1) {
-			throw new InvalidPresenterException("Multiple services of type $class found: " . implode(', ', $services) . '.');
-
-		} elseif (!$services) {
-			if ($this->touchToRebuild) {
-				touch($this->touchToRebuild);
-			}
-
-			$presenter = $this->container->createInstance($class);
-			$this->container->callInjects($presenter);
-			if ($presenter instanceof UI\Presenter && $presenter->invalidLinkMode === NULL) {
-				$presenter->invalidLinkMode = $this->container->parameters['debugMode'] ? UI\Presenter::INVALID_LINK_WARNING : UI\Presenter::INVALID_LINK_SILENT;
-			}
-			return $presenter;
-		}
-
-		return $this->container->createService($services[0]);
+		return call_user_func($this->factory, $this->getPresenterClass($name));
 	}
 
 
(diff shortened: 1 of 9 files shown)


src/Application/PresenterFactory.php
```diff
@@ -35,11 +35,15 @@ class PresenterFactory extends Nette\Object implements IPresenterFactory
...
+	/** @var int */
+	private $invalidLinkMode;
 
-	public function __construct(Nette\DI\Container $container, $autoRebuild = FALSE)
+
+	public function __construct(Nette\DI\Container $container, $autoRebuild = FALSE, $invalidLinkMode = NULL)
 	{
 		$this->container = $container;
 		$this->autoRebuild = $autoRebuild;
+		$this->invalidLinkMode = $invalidLinkMode;
```
fprochazka: I liked more the setter. A lot of people are inheriting the class and this might make an unnecesary BC Break. It would be imho better to add sane default value to the property and create setter.

klimesf:
:+1:

src/Bridges/ApplicationDI/Helpers.php
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @internal
+ */
+class Helpers
+{
+
+	/**
+	 * @return Nette\Application\IPresenter
+	 */
+	public static function createPresenter($class, $container, $touchToRebuild, $invalidLinkMode)
```
stekycz: Typehint for `$container`?

What does `$touchToRebuild` mean? What should be the value? A comment would be nice.

dg: I added typehint.

hrach:
- first commit contains changes with invalid link hadling, probably bug
- at least invalid link handling should be postponed to 2.4

-----

PresenterFactory:
- current proposed design doesn't make sense at all,
- there is nothing wrong to have a dependency on DIC, if there is a interface and possibility for own replacement.
- create Factory, which in fact deosn't create a instance, is wrong. Passing logic through clousure is wrong. In the end, the dependency on the DIC is still present, in static Helper.
- Corrent solution should be:
  - provide default implementation without DIC.
  - provide DIC implementation in bridges namespace.

src/Bridges/ApplicationDI/ApplicationExtension.php
```diff
@@ -56,10 +64,16 @@ public function loadConfiguration()
 			$application->addSetup('Nette\Bridges\ApplicationTracy\RoutingPanel::initializePanel');
 		}
 
-		$touchToRebuild = $this->debugMode && $config['scanDirs'] ? reset($config['scanDirs']) : NULL;
+		$code = '!($me = $this) ?: ' . $container->formatPhp(
+			'function ($class) use ($me) { return Nette\Bridges\ApplicationDI\Helpers::createPresenter($me, $class, ?, ?); }',
```
fprochazka: I like the closure, but I don't like that it's static class, when it doesn't have to be static class.

you could have created

```php
class PresenterFactoryCallback
{
    public function __construct(/* deps */);
    public function __invoke($class);
}
```

and attach it like

```php
$presenterFactory = $container->addDefinition($this->prefix('presenterFactory'))
    ->setFactory('Nette\Application\PresenterFactory', [
        new Nette\DI\Statement('PresenterFactoryCallback')
    ])
```

IMHO a lot nicer, what do you think?

dg:
@hrach  I see no changes in invalid link handling in first commit.

This can be of course postponed to 2.4, but it seems to me unnecessary to wait a year before invalid links will throw warnings. 

ad design: to create interface & its implementation where can be used simple callback is possible, but over complicated.

fprochazka:
I'm :+1: for warnings on invalid links. They don't kill production and I wanna know about them.

hrach:
@dg sorry, I missread.

> ad design: to create interface & its implementation where can be used simple callback is possible, but over complicated.

That's more Filip's proposal than mine. I see no reason to create another interface, when one is present. It's IPresenterFactory. It's totally ok to have one imlementaion dependent on DIC:

src/Bridges/ApplicationDI/ApplicationExtension.php
dg: Yes, it is fine. I take this more as a cosmetic detail, which can be changed at any time.

dg:
>  It's totally ok to have one imlementaion dependent on DIC:

I agree, the motivation was to get rid of frequent changes in constructor, https://github.com/nette/application/pull/66#discussion_r25307535

PR #67, opened 2015-03-03
composer.json: requires RobotLoader ~2.2

enumag:
ApplicationExtension doesn't work without RobotLoader. (Unless you disable scanDirs option.)


dg:
This is bridge dependency and these dependencies are not managed by composer.json.

enumag:
In that case why is nette/di required?

dg:
Because it was used by PresenterFactory. Now it can be moved to require-dev.

PR #71, opened 2015-03-18
Proper enconding of Content-Dispisition filename

foxycode:
According to RFC 5987 Content-Disposition filename should not contain non-US-ASCII characters.

Proper way to encode filename is this:
"filename*=utf-8''%C5%BElu%C5%A5ou%C4%8Dk%C3%BD%20k%C5%AF%C5%88.txt".

Unfortunately, RFC 5987 is not supported by all browsers, so for compatibility reasons, filename should be sent like this:
"filename=žluťoučký kůň.txt; filename*=utf-8''%C5%BElu%C5%A5ou%C4%8Dk%C3%BD%20k%C5%AF%C5%88.txt".

Some discussion about this is here: http://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http

### DIFF
--- a/src/Application/Responses/FileResponse.php
+++ b/src/Application/Responses/FileResponse.php
@@ -93,7 +93,9 @@ class FileResponse extends Nette\Object implements Nette\Application\IResponse
 	{
 		$httpResponse->setContentType($this->contentType);
 		$httpResponse->setHeader('Content-Disposition',
-			($this->forceDownload ? 'attachment' : 'inline') . '; filename="' . $this->name . '"');
+			($this->forceDownload ? 'attachment' : 'inline')
+				. '; filename="' . $this->name . '"'
+				. '; filename*=utf-8\'\'' . rawurlencode($this->name));
 
 		$filesize = $length = filesize($this->file);
 		$handle = fopen($this->file, 'r');
--- a/tests/Application/FileResponse.contentDisposition.phpt
+++ b/tests/Application/FileResponse.contentDisposition.phpt
@@ -28,7 +28,7 @@ test(function() {
 	$fileResponse->send(new Http\Request(new Http\UrlScript), $response = new Http\Response);
 
 	Assert::same( $origData, ob_get_clean() );
-	Assert::same( 'attachment; filename="' . $fileName . '"', $response->getHeader('Content-Disposition') );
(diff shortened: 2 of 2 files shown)


fprochazka:
Great work :+1:

src/Application/Responses/FileResponse.php
```diff
@@ -93,7 +93,9 @@ public function send(Nette\Http\IRequest $httpRequest, Nette\Http\IResponse $htt
 	{
 		$httpResponse->setContentType($this->contentType);
 		$httpResponse->setHeader('Content-Disposition',
-			($this->forceDownload ? 'attachment' : 'inline') . '; filename="' . $this->name . '"');
+			($this->forceDownload ? 'attachment' : 'inline')
+				. '; filename="' . $this->name . '"'
+				. '; filename*=utf-8\'\'' . rawurldecode($this->name));
```
dg: decode or encode?

foxycode:
@dg Of course encode :-) Updated.

dg:
Should it be `"filename=žluťoučký kůň.txt; filename*=utf-8''%C5%BElu%C5%A5ou%C4%8Dk%C3%BD%20k%C5%AF%C5%88.txt"` or `"filename=zlutoucky kun.txt; filename*=utf-8''%C5%BElu%C5%A5ou%C4%8Dk%C3%BD%20k%C5%AF%C5%88.txt"`? (i.e. in ASCII)

foxycode:
@dg I tried current approach with `"filename="žluťoučký kůň.txt"; filename*=utf-8''%C5%B...` and it works great in FF 36, Chrome 41 and IE 11. According linked discussion, browsers without RFC 5987 support should ignore second filename part and thus use same filename which nette sent before.

dg:
Can you add prefix `FileResponse:` in commit message?

dg:
Btw Test Cases for HTTP Content-Disposition header field http://greenbytes.de/tech/tc2231/

foxycode:
Prefix added, I'll look to your link more tomorrow, but according to what I've read my solution should work just fine.

dg:
Thanks!

foxycode:
@dg When you plan next release (2.3.2)?

dg:
Soon

PR #79, opened 2015-06-07
Application: Availability of invalid request for ErrorPresenter

enumag:
Currently if the initial request is invalid, there is no way for the ErrorPresenter to know what the initial request was. Or in case there is a [mechanic to use per-module ErrorPresenters](https://github.com/enumag/Application/blob/master/src/Application.php#L58-L90) it will fail to call the correct module-specific ErrorPresenter.

For these reasons I believe that the initial request should be available even if it is not valid.

(Just to be clear, I'm not looking for hack suggestions. I'm well aware I can hack it using `$request = $this->router->match($this->httpRequest);` in case the array is empty or just override the createInitialRequest method.)

### DIFF
--- a/src/Application/Application.php
+++ b/src/Application/Application.php
@@ -117,14 +117,14 @@ class Application extends Nette\Object
 			throw new BadRequestException('No route for HTTP request.');
 
 		} elseif (strcasecmp($request->getPresenterName(), $this->errorPresenter) === 0) {
-			throw new BadRequestException('Invalid request. Presenter is not achievable.');
+			throw BadRequestException::fromRequest($request, 'Invalid request. Presenter is not achievable.');
 		}
 
 		try {
 			$name = $request->getPresenterName();
 			$this->presenterFactory->getPresenterClass($name);
 		} catch (InvalidPresenterException $e) {
-			throw new BadRequestException($e->getMessage(), 0, $e);
+			throw BadRequestException::fromRequest($request, $e->getMessage(), 0, $e);
 		}
 
 		return $request;
@@ -169,7 +169,7 @@ class Application extends Nette\Object
 			$this->httpResponse->setCode($e instanceof BadRequestException ? ($e->getCode() ?: 404) : 500);
 		}
 
-		$args = ['exception' => $e, 'request' => end($this->requests) ?: NULL];
+		$args = ['exception' => $e, 'request' => $e->getRequest() ?: (end($this->requests) ?: NULL)];
 		if ($this->presenter instanceof UI\Presenter) {
 			try {
 				$this->presenter->forward(":$this->errorPresenter:", $args);
--- a/src/Application/exceptions.php
+++ b/src/Application/exceptions.php
@@ -44,9 +44,29 @@ class BadRequestException extends \Exception
 	protected $code = 404;
 
 
-	public function __construct($message = '', $code = 0, \Exception $previous = NULL)
+	/** @var Request */
+	private $request;
+
+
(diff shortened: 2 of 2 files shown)


dg:
This solution seems like unexpected side effect of method named createXyz()

JanTvrdik:
What about putting the `$request` instance inside the `BadRequestException`?

enumag:
@JanTvrdik Yeah, that's the other option I was considering. Look at my implementation please.

src/Application/Application.php
```diff
@@ -117,14 +117,14 @@ public function createInitialRequest()
...
-			throw new BadRequestException('Invalid request. Presenter is not achievable.');
+			throw (new BadRequestException('Invalid request. Presenter is not achievable.'))->setRequest($request);
 		}
 
 		try {
 			$name = $request->getPresenterName();
 			$this->presenterFactory->getPresenterClass($name);
 		} catch (InvalidPresenterException $e) {
-			throw new BadRequestException($e->getMessage(), 0, $e);
+			throw (new BadRequestException($e->getMessage(), 0, $e))->setRequest($request);
```
fprochazka: doctrine style

```php
throw BadRequestException::fromRequest($request, $e);
```

would be nicer :P

JanTvrdik: :+1: for Doctrine style. Or we can at least remove `setRequest()` and pass it as 4th constructor's parameter. Exceptions should be immutable.

enumag: Thanks for suggestions! :-) I used both.

src/Application/exceptions.php
```diff
@@ -44,9 +44,29 @@ class BadRequestException extends \Exception
...
+	/**
+	 * @return Request|null
+	 */
+	public function getRequest()
+	{
+		return $this->request;
+	}
+
+
+	public static function fromRequest(Request $request, $message = '', $code = 0, \Exception $previous = NULL)
```
enumag: Maybe just two parameters where the second could be either string or \Exception? cc @fprochazka

fprochazka: I was thinking `public static function fromRequest(Request $request, \Exception $previous = NULL)` since it doesn't obstruct the original constructor and you can still overwrite it

enumag: I think you missed that the first use-case in Appliction does not have an original exception.

fprochazka: I see now, thanks :)

enumag: @fprochazka Then what do you think is better? The implementation currently proposed or this one? :-)

```php
/**
 * @param string|\Exception $message
 */
public static function fromRequest(Request $request, $message)
{
	$cond = $message instanceof \Exception;
	return new static($cond ? $message->getMessage(), 0, $cond ? $message : null, $request);
}
```

enumag:
By the way what do you think about this commit? https://github.com/enumag/application-1/commit/3f5307a5a383710ec8fce2cd23aadf1ed3a35410 @fprochazka @JanTvrdik

enumag:
ping @fprochazka @JanTvrdik

enumag:
bump

enumag:
ping @dg

dg:
IMHO the 'there is no way for the ErrorPresenter to know what the initial request was' was fixed in 7b07c4e3aa150fdea0b7765a7223bca621136689, ie. now it can be received via `$application->getRequests()`

enumag:
Thanks!

PR #80, opened 2015-06-12
Presenter: removed static from refUrl variable

Yrwein:
- bugfix
- documentation - not needed
- BC break - no

The pull request fixes issue when testing presenters without isolation: You make two presenter calls with Nette\Http\Request-s on different domains - the first domain is cached for whole class (and not only instance) and so the second instance of the same presenter will redirect you to the first domain.
The test is appended mainly as example (there's no problem to remove it if it's not needed).

### DIFF
--- a/src/Application/UI/Presenter.php
+++ b/src/Application/UI/Presenter.php
@@ -126,6 +126,9 @@ abstract class Presenter extends Control implements Application\IPresenter
 	/** @var ITemplateFactory */
 	private $templateFactory;
 
+	/** @var Nette\Http\Url */
+	private $refUrlCache;
+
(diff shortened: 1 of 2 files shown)


fprochazka:
:+1: I like it

Yrwein:
Frankly, I don't think there will be measurable performance hit from changing function's static variable to property. Or is there something I'm missing? (Nevertheless we can wait few weeks until it's deployed.)

rostenkowski:
Hi there:) 

what's the current status of this pull request?

rostenkowski:
Thank you 
:+1:

PR #83, opened 2015-06-29
Class Link is no longer marked as internal.

foxycode:


dg:
Commit message :-(((

foxycode:
@dg I don't know what else to write, you can tell me and I update it.

dg:
Just try to be consistent with other messages

foxycode:
@dg Hope this is ok.

JanTvrdik:
Much better! To make it perfect it should not end with full stop and instead the word "flag" I would you "annotation".

foxycode:
@JanTvrdik done

dg:
:+1:

PR #87, opened 2015-07-28
ApplicationExtension should not disable autowiring of general IPresenter service

JanTvrdik:
ApplicationExtension currently [disables autowiring](https://github.com/nette/application/blob/1ecd8ed01e369f183150a171d949e4e25a22967f/src/Bridges/ApplicationDI/ApplicationExtension.php#L109) of all services which implement `IPresenter`. This seems like too aggressive decision, it should only be done for `UI\Presenter` descendants.

cc @mikulas

### DIFF
--- a/src/Bridges/ApplicationDI/ApplicationExtension.php
+++ b/src/Bridges/ApplicationDI/ApplicationExtension.php
@@ -106,9 +106,9 @@ class ApplicationExtension extends Nette\DI\CompilerExtension
 		}
 
 		foreach ($all as $def) {
-			$def->setInject(TRUE)->setAutowired(FALSE)->addTag('nette.presenter', $def->getClass());
+			$def->setInject(TRUE)->addTag('nette.presenter', $def->getClass());
 			if (is_subclass_of($def->getClass(), 'Nette\Application\UI\Presenter')) {
-				$def->addSetup('$invalidLinkMode', [$this->invalidLinkMode]);
+				$def->setAutowired(FALSE)->addSetup('$invalidLinkMode', [$this->invalidLinkMode]);
 			}
 		}
 	}


JanTvrdik:
@dg Can you think of any problem with having custom implementations of `IPresenter` with enabled autowiring?

JanTvrdik:
Note: we may also consider calling `setInject` only for `UI\Presenter` descendants (because users can enable inject with decorator if they need it), but having inject enabled by default does not cause me any trouble.

JanTvrdik:
ping @dg

dg:
1) setInject only for UI\Presenter - makes sense
2) should not disable autowiring of general IPresenter service - not makes sense. At least it should disable autowiring for all automatically added presenters.

JanTvrdik:
> At least it should disable autowiring for all automatically added presenters

That would probably make even less sense. Either all presenters must always be NOT autowired OR there must be a rule for why some presenter are autowired and some are not (regardless of how they are registered in DI container). Or I could be wrong, obviously.

---

Let me write some random arguments I can think of for both sides:

* I can argue that `IPresenter` interface design is autowiring-friendly (e.g. it does not accept `Application\Request` in constructor. From `IPresenter` interface point of view there is no reason to not be autowired.
* You can argue that `IPresenter` interface exists together with `IPresenterFactory` which is the ONLY valid way to create a presenter instance.
* I can argue that I don't see why `IPresenterFactory` should be the only way and that we should consider redesigning the `IPresenterFactory` (or its implementation) to support autowired presenters (to reuse existing instance).
* You can argue that there is a theoretical security risk with using an already existing instance of presenter because programmer may have not properly reset the object state.
* I can argue that if someone is smart enough to implement custom `IPresenter` implementation he should be smart enough to take care of it. Also even with autowired presenters reusing existing instance will require modifying `IPresenterFactory` as well which makes it supper unlikely.
* You can argue that the autowire inconsistency (every class should either be autowired or have factory, never both) with `IPresenterFactory` is really annoying.
* And I would agree that is too general statement which you cannot prove although I find the inconsistency annoying as well (meaning I have not yet came up with a mental model where it would make sense).

----

What I have miss?

dg:
You should never get presenters via `getService()` or `getByType()`. Currently, it is not possible to completely forbid it, only autowiring can be disabled.

JanTvrdik:
But why? Image a world without UI\Presenter. What would be the reason for not autowiring presenters?

jkuchar:
I also do not get, why it is wrong to not autowire presenters.

JanTvrdik:
ping, this is still not resolved

dg:
Current nette/application is simply not designed for reusing presenters, so there are names like  `PresenterFactory::createPresenter` instead of (for example) `PresenterLocator::getPresenter()`.

At the same time, nette/application is designed to work with set of presenters (therefore mandatory argument `$name` in `createPresenter`).

Of couse some of them can be reused, i.e. `NetteModule\ErrorPresenter`. And it makes sense to receive it via `getService('NetteModule\ErrorPresenter')`.

But it is wrong to receive them via `getService('Nette\Application\IPresenter')`.

With current nette/di it can be reached by:
- disabling autowiring only for UI\Presenter as you suggest
- and adding IPresenter to excluded classes

PR #88, opened 2015-07-29
Echo return of control render methods

MartinMystikJonas:
See http://forum.nette.org/cs/17928-rfc-komponenty-ktere-se-samy-nevykresluji-ale-jen-vraceji-view#p160254

This change will allow component to instead of directly render itslef to return object representing component view. This view than can by examined in tests before it is rendered. 

Render methods does not return anything yet so this sould not be BC break.

### DIFF
--- a/src/Bridges/ApplicationLatte/UIMacros.php
+++ b/src/Bridges/ApplicationLatte/UIMacros.php
@@ -76,7 +76,7 @@ if (empty($_l->extends) && !empty($_control->snippetMode)) {
 		return ($name[0] === '$' ? "if (is_object($name)) \$_l->tmp = $name; else " : '')
 			. '$_l->tmp = $_control->getComponent(' . $name . '); '
 			. 'if ($_l->tmp instanceof Nette\Application\UI\IRenderable) $_l->tmp->redrawControl(NULL, FALSE); '
-			. ($node->modifiers === '' ? "\$_l->tmp->$method($param)" : $writer->write("ob_start(); \$_l->tmp->$method($param); echo %modify(ob_get_clean())"));
+			. ($node->modifiers === '' ? "echo \$_l->tmp->$method($param)" : $writer->write("ob_start(); echo \$_l->tmp->$method($param); echo %modify(ob_get_clean())"));
 	}
 
 


src/Bridges/ApplicationLatte/UIMacros.php
```diff
@@ -76,7 +76,7 @@ public function macroControl(MacroNode $node, PhpWriter $writer)
 		return ($name[0] === '$' ? "if (is_object($name)) \$_l->tmp = $name; else " : '')
 			. '$_l->tmp = $_control->getComponent(' . $name . '); '
 			. 'if ($_l->tmp instanceof Nette\Application\UI\IRenderable) $_l->tmp->redrawControl(NULL, FALSE); '
-			. ($node->modifiers === '' ? "\$_l->tmp->$method($param)" : $writer->write("ob_start(); \$_l->tmp->$method($param); echo %modify(ob_get_clean())"));
+			. ($node->modifiers === '' ? "\$_l->tmp->$method($param)" : $writer->write("ob_start(); echo \$_l->tmp->$method($param); echo %modify(ob_get_clean())"));
```
mishak87: Output buffer is not necessary:
```php
echo %modify(\$_l->tmp->$method($param));
```

MartinMystikJonas: I was not sure why it is there so I keep it. Why it was there in first place?

MartinMystikJonas: It was added by commit: Implemented possibility of helpers for "control" macro https://github.com/nette/nette/commit/aa9ee5bcf28cced776485df10a5a58fa31508257

dg:
It is unneccessary complication.

MartinMystikJonas:
@dg: Well it would greatly siplified testing of components. Is there any reason to not do this?

dg:
It will not because you will check both, output and returned value, instead of only output.

MartinMystikJonas:
For components which use returning of view it's not necessary to check output. Simplification of testing is in shift from parsing text output to checking structured data (returned from render).

For more complex components if would be much easier to just check values passed to template instead trying to find relevant pieces of data in generated output.

fprochazka:
@MartinMystikJonas the exclusivity thing would be harder to achieve imho.

MartinMystikJonas:
@fprochazka Well it would require use of output buffer. Which could be performace problem in some cases.

dg:
Testing is so easy, if the point is easier testing, it cannot compensate bigger complexity.

PR #90, opened 2015-08-23
5.6 with ::class

JanTvrdik:


dg:
@JanTvrdik thanks for all these PR!

PR #91, opened 2015-09-07
Removing nette/reflection [WIP]

dg:


src/Application/UI/Presenter.php
```diff
@@ -504,7 +504,7 @@ public function formatLayoutTemplateFiles()
 		$name = $this->getName();
 		$presenter = substr($name, strrpos(':' . $name, ':'));
 		$layout = $this->layout ? $this->layout : 'layout';
-		$dir = dirname($this->getReflection()->getFileName());
+		$dir = dirname((new \ReflectionClass($this))->getFileName());
```
Majkl578: Why not ReflectionObject?

dg: Why ReflectionObject?

Majkl578: Because `$this` is an instance, not a type name.

dg: It works, ReflectionClass accepts objects.

Majkl578: But ReflectionObject is semantically more correct, and also extends ReflectionClass.

enumag: What's the difference between ReflectionClass and ReflectionObject anyway? I couldn't find anything worth mentioning in the docs.

dg: I think that ReflectionObject::getProperties returns real object properties and ReflectionClass::getProperties returns default properties from a class.

enumag:
@dg What about composer.json?

dg:
In next version.

Majkl578:
Why? It's not needed by this package so it should not be there, only in suggests section. If anyone was using it directly, they've been shooting themselves in the foot -- they should've had it in their composer.json.

dg:
Because back compatibility

Majkl578:
I don't get it. This package doesn't need that package anymore so it should just drop the dependency entirely. Noone should ever depend on a library indirectly through another dependency because it is just not what _dependency_ means.
Also it's ok to remove nette/security, but not nette/reflection? Doesn't make sense.

dg:
It is directly used in code.

enumag:
Even if this affects anyone its easily fixable by one line in composer.json. Its ok to put a minor BC break like that into 2.4 in my opinion.

PR #92, opened 2015-09-15
link generator: fixed leading colon [closes #72]

### LINKED ISSUE #72: Leading colon in the destination breaks link generation using LinkGenerator
We use modules in our application and sometimes we need to specify the destination absolutely using the leading colon. Something like `:Module:Presenter:action`. This works for normal link generation using `$presenter->link()` (or in templates) but the brand new `LinkGenerator` throws an `InvalidLinkException`  with the message `Presenter name must be alphanumeric string, ':Module:Presenter' is invalid`.

If I remove the leading colon, the `LinkGenerator` starts working, but now it's `$presenter->link()` that fails with an exception.

I might be doing something wrong but I'm pretty sure that the behavior should be consistent.

ondrakub:


fprochazka:
The link is always absolute. Wouldn't it be better to throw exceptions?

janedbal:
@fprochazka Leading colon is not used for absolute link, but for referencing different module.

fprochazka:
@janedbal Not exactly. In presenters, you always write relative paths, but there is no `../`, so `:` exists and therefore you can write absolute path to module. In LinkGenerator, there are no relative links. Therefore makes no sense a syntax for absolute paths.

fprochazka:
@janedbal I wrote "link" not a "URL". In presenters, you always write relative links, but there is no `../`, so `:` exists and therefore you can write absolute links to module. In LinkGenerator, there are no relative links. Therefore makes no sense a syntax for absolute links - it's always absolute.

The question is.. should we ignore the colon for compatibility's sake, or forbit it because it makes no sense?

Majkl578:
This is pretty much the same as prefixing namespaces with `\`. Would you throw an exception for that or use ltrim?

fprochazka:
I'm not sure :)

JanTvrdik:
For compatibility with presenters you would need to NOT support links without the leading colon as the meaning may now be different.

I'm :-1: on this.

hrach:
This is not about compatibility but usability. Neon also accepts FQN with the leading slash, although it requires FQN by design.

dg:
Once someone asked me: what is the difference between `{block #name}` and `{block name}`? I replied that there is no difference. And he asked me „So why there are two ways to do one thing?“ 

There should be one-- and preferably only one --obvious way to do it. https://www.python.org/dev/peps/pep-0020/

Majkl578:
> There should be one-- and preferably only one --obvious way to do it.

:+1: But as with namespaces and FQCN `\` prefix, even though leading `\` does not make any sense e.g. in configuration, people are lazy and don't want to care, especially if the meaning is the same.

hrach:
@dg

- neon has two options; this is much more similar to the namespace case than to the block case.
- if you are linking zen of python, you probably know, that you are breaking the second rule: "Explicit is better than implicit.".

JanTvrdik:
> neon has two options

Maybe we should remove one.

PR #93, opened 2015-09-21
added some annotations

pavelvondrasek:
added annotation for Nette\Application\UI\Presenter::$template property, fixed "Field declared dynamically" in PhpStorm

added annotation for Nette\Application\UI\Multiplier::__construct method, fixed "Expected Nette\ComponentModel\IContainer got \Closure" in PhpStorm

### DIFF
--- a/src/Application/UI/Multiplier.php
+++ b/src/Application/UI/Multiplier.php
@@ -19,6 +19,9 @@ class Multiplier extends PresenterComponent
 	private $factory;
 
 
+	/**
+	 * @param callable $factory
+	 */
 	public function __construct(callable $factory)
 	{
 		parent::__construct();
--- a/src/Application/UI/Presenter.php
+++ b/src/Application/UI/Presenter.php
@@ -21,6 +21,7 @@ use Nette\Http;
  * @property-read string $action
  * @property      string $view
  * @property      string $layout
+ * @property-read ITemplate|\stdClass $template
  * @property-read \stdClass $payload
  * @property-read bool $ajax
  * @property-read Nette\Application\Request $lastCreatedRequest


src/Application/UI/Multiplier.php
```diff
@@ -19,6 +19,9 @@ class Multiplier extends PresenterComponent
 	private $factory;
 
 
+	/**
```
dg: This is not needed

src/Application/UI/Presenter.php
```diff
@@ -21,6 +21,7 @@
  * @property-read string $action
  * @property      string $view
  * @property      string $layout
+ * @property-read ITemplate|\stdClass $template
```
dg: This is declared on Control.

dg:
Open issue here https://youtrack.jetbrains.com/issues/WI

src/Application/UI/Presenter.php
racinmat: But in control it is not \stdClass, so PhpStorm shows every declaration of template variable as warning.

src/Application/UI/Multiplier.php
racinmat: When it is not here, PhpStorm shows "Expected Nette\ComponentModel\IContainer got \Closure" warning. And it marks code as wrong, which is very unpleasant, since properly working code should not be marked as bad one.

matej21: IMHO it is phpstorm bug

pavelvondrasek: ok

src/Application/UI/Presenter.php
pavelvondrasek: But in UI\Control is these declaration: @property-read ITemplate $template

But PhpStorm cannot detect that $template type is \stdClass

pavelvondrasek:
OK, thank you. :)

src/Application/UI/Presenter.php
enumag: Template is not stdClass.

PR #94, opened 2015-09-30
Router: Have a routes collection interface & support in Tracy router panel

Andrewsville:
* Feature
* Documentation – No change needed, unless you want to mention it in the "Custom Router" chapter.
* BC break - no

Currently, the only way to have a collection of routes and have them displayed in the Tracy panel is to use the ```RouteList``` class.

This PR generalizes this logic to allow developers to implement their own route collections maybe with some logic inside (the ```RouteList``` is just a simple hashmap), maybe immutable (```RouteList``` is mutable), etc. It comes from our use case but I believe it could be useful for others, too.

* This PR introduces a new interface for route lists that extends ```IteratorAggregate``` (which is the easiest way to implement a ```Traversable```) and adds the ```getModule()``` method that is required by the Tracy panel.
* Because of the way ```RouteList``` is implemented, it can implement this new interface without any changes.
* And the last change is the ```instanceof``` typecheck in the Panel to display a routes collection.

### DIFF
--- a/src/Application/IRouteList.php
+++ b/src/Application/IRouteList.php
@@ -1,0 +1,23 @@
+<?php
+
+/**
+ * This file is part of the Nette Framework (http://nette.org)
+ * Copyright (c) 2004 David Grudl (http://davidgrudl.com)
+ */
+
+namespace Nette\Application;
+
+use Nette;
+
+
+/**
+ * Routes collection.
+ */
+interface IRouteList extends \IteratorAggregate
+{
+	/**
+	 * @return string
+	 */
+	function getModule();
+
+}
--- a/src/Application/Routers/RouteList.php
+++ b/src/Application/Routers/RouteList.php
@@ -13,7 +13,7 @@ use Nette;
 /**
  * The router broker.
  */
-class RouteList extends Nette\Utils\ArrayList implements Nette\Application\IRouter
+class RouteList extends Nette\Utils\ArrayList implements Nette\Application\IRouter, Nette\Application\IRouteList
 {
 	/** @var array */
 	private $cachedRoutes;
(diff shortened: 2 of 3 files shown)


pavelkouril:
Personally, I like this idea. :) :+1:

src/Application/IRouteList.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Nette\Application;
+
+use Nette;
+
+
+/**
+ * Routes collection.
+ */
+interface IRouteList extends \IteratorAggregate
```
JanTvrdik: Why not extend Traversable?

JanTvrdik: Why not extend `IRouter`?

JanTvrdik:
I dislike the `getModule()` method on`IRouteList`.

src/Application/IRouteList.php
Majkl578: User-land code can't extend Traversable directly.

dg: Interface can extend, but class can't implement.

Andrewsville:
Well, it is required by the Panel. There's not much I can do about it.

src/Application/IRouteList.php
dg: This is funny:

this works:
```php
interface A extends Traversable 
{}

class B implements IteratorAggregate, A
{
	function getIterator() {}
}
```

this triggers Fatal error: Class B must implement interface Traversable as part of either Iterator or IteratorAggregate:

```php
interface A extends Traversable 
{}

class B implements A, IteratorAggregate
{
	function getIterator() {}
}
```

Andrewsville: I have encountered this (the need to have interfaces in the correct order) when developing the PR and it is exactly the reason I won't implement it using Traversable.

dg: It is documented http://php.net/class.traversable

f3l1x:
I like it. Could we merge it, rebase it or close it?

PR #100, opened 2015-10-13
PresenterComponentReflection::combineArgs() should retype variables i…

### LINKED ISSUE #99: PresenterComponentReflection::combineArgs() doesn't properly assign retyped variables
combineArgs() retypes values, but if they are scalars, they are not changed in the returned array

Broken by:
https://github.com/nette/application/commit/e4eb640294e8091d6a735bb02f12aef00457be30#diff-07e4bc94262f0f87556bb161b6386767L128

fix incoming

peteruhnak:
…n the $res array. [Fixes #99]

### DIFF
--- a/src/Application/UI/PresenterComponentReflection.php
+++ b/src/Application/UI/PresenterComponentReflection.php
@@ -117,15 +117,14 @@ class PresenterComponentReflection extends Nette\Reflection\ClassType
 	public static function combineArgs(\ReflectionFunctionAbstract $method, $args)
 	{
 		$res = [];
-		$i = 0;
-		foreach ($method->getParameters() as $param) {
+		foreach ($method->getParameters() as $i => $param) {
 			$name = $param->getName();
 			if (!isset($args[$name]) && $param->isDefaultValueAvailable()) {
-				$res[$i++] = $param->getDefaultValue();
+				$res[$i] = $param->getDefaultValue();
 			} else {
-				$res[$i++] = $arg = isset($args[$name]) ? $args[$name] : NULL;
+				$res[$i] = $arg = isset($args[$name]) ? $args[$name] : NULL;
 				list($type, $isClass) = self::getParameterType($param);
-				if (!self::convertType($arg, $type, $isClass)) {
+				if (!self::convertType($res[$i], $type, $isClass)) {
 					throw new BadRequestException(sprintf(
 						'Argument $%s passed to %s() must be %s, %s given.',
 						$name,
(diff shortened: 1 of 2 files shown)


dg:
Nice catch!

tests/UI/PresenterComponentReflection.combineArgs.phpt
```diff
@@ -0,0 +1,24 @@
...
+class MyPresenter
+{
+
+	public function myMethod($intParam = 0, $strParam = '')
+	{
+	}
+
+}
+
+$reflection = Nette\Utils\Callback::toReflection(Nette\Utils\Callback::check(['MyPresenter', 'myMethod']));
```
dg: Can you change it to simpler `$reflection = new ReflectionMethod('MyPresenter', 'myMethod');` ?

hranicka:
:+1: 

It's an annoying bug in v2.3.6.

src/Application/UI/PresenterComponentReflection.php
```diff
@@ -125,7 +125,7 @@ public static function combineArgs(\ReflectionFunctionAbstract $method, $args)
 			} else {
 				$res[$i++] = $arg = isset($args[$name]) ? $args[$name] : NULL;
```
enumag: `$res[$i]`

src/Application/UI/PresenterComponentReflection.php
```diff
@@ -125,7 +125,7 @@ public static function combineArgs(\ReflectionFunctionAbstract $method, $args)
...
 				list($type, $isClass) = self::getParameterType($param);
-				if (!self::convertType($arg, $type, $isClass)) {
+				if (!self::convertType($res[$i - 1], $type, $isClass)) {
```
enumag: `$res[$i++]`

src/Application/UI/PresenterComponentReflection.php
peteruhnak: How about I use the foreach instead?

```
foreach ($method->getParameters() as $i => $param) {
```

enumag: If the items in `$method->getParameters()` are indexed by numbers starting from zero then yes, it would be better.

peteruhnak: Yeah, it returns regular array indexed from zero; I've changed it.

enumag:
@peteruhnak Squash all the commits into one please.

hranicka:
@dg Please when we would expect fixed release v2.3.7 (for orientation)?

dg:
Soon.

PR #102, opened 2015-10-24
Route: fix old Nette\Callback name

TomasVotruba:
New namespace is `Nette\Utils\Callback`.


dg:
This is for compatibility with old name.

TomasVotruba:
I see. What about using to both to make this clear in the code as well?

dg:
Nette\Utils\Callback is static class.

TomasVotruba:
I don't understand.

dg:
Me too ;)

enumag:
@TomasVotruba An instance of Nette\Utils\Callback is NOT callable.

TomasVotruba:
@enumag I see. They're different.

That's settled then.

PR #103, opened 2015-11-06
Added support for request restore with ajax

sallyx:
Application: added support for request restore with ajax


dg:
How many times do I have to repeat „writes a consistent commit message“? I'm tired.

Tests fails.

sallyx:
I have cloned master branch, where tests fails too. As soon as the master brnach will be ok, I will merge it ...

dg:
Tests fails after your commit, see https://travis-ci.org/nette/application/builds/87526325

PR #104, opened 2015-11-09
Template: better error message when Translator is not set

greeny:
Ref: https://forum.nette.org/en/24756-unclear-error-message-when-translator-is-not-set-to-template

### DIFF
--- a/src/Bridges/ApplicationLatte/TemplateFactory.php
+++ b/src/Bridges/ApplicationLatte/TemplateFactory.php
@@ -84,6 +84,12 @@ class TemplateFactory extends Nette\Object implements UI\ITemplateFactory
 			return $time == NULL ? NULL : Nette\Utils\DateTime::from($time)->modify($delta . $unit); // intentionally ==
 		});
 
+		if (!isset($latte->getFilters()['translate'])) {
+			$latte->addFilter('translate', function () {
+				throw new Nette\InvalidStateException('Translator has not been set. Set translator using $template->setTranslator().');
+			});
+		}
+
 		// default parameters
 		$template->control = $template->_control = $control;
 		$template->presenter = $template->_presenter = $presenter;


src/Bridges/ApplicationLatte/TemplateFactory.php
```diff
@@ -83,6 +83,12 @@ public function createTemplate(UI\Control $control = NULL)
 		$latte->addFilter('modifyDate', function ($time, $delta, $unit = NULL) {
 			return $time == NULL ? NULL : Nette\Utils\DateTime::from($time)->modify($delta . $unit); // intentionally ==
 		});
+		$latte->addFilter('translate', function ($text/*, ...$params*/) use ($template) {
+			if (($translator = $template->getTranslator()) && $translator instanceof Nette\Localization\ITranslator) {
+				return call_user_func_array([$translator, 'translate'], func_get_args());
+			}
+			throw new Nette\InvalidStateException('Translator has not been set.');
+		});
```
JanTvrdik: * master is PHP 5.6+, so you use `...$params`
* the `instanceof` check is useless, because `setTranslator` already handles this
* translate filter is **performance critical** (because it can be called 100×/request), this change makes it a lot slower, you would need to extract the check outside of translate call

greeny: **ad. 1st && 2nd point**: Fixed
**ad. 3rd point**: I agree with that, but when should be this check executed? You can never know when `ITranslator` will be passed to `Template`.

My only idea about this is something simlar like:

```php
$latte->addFilter('translate', function ($text, ...$params) use ($latte, $template) {
	if ($translator = $template->getTranslator()) {
		$latte->addFilter('translate', function ($text, ...$params) use ($template) {
			return call_user_func_array([$translator, 'translate'], func_get_args());
		}
		return $template->translate($text, ...$params));
	}
	throw new Nette\InvalidStateException('Translator has not been set.');
});

```

This breaks in situations, where you set different `Translator` in the middle of running app (but I think thats ok, I cannot think about any use-case)

matej21: What about this?
```php
if (!isset($latte->getFilters()['translate'])) {
    $latte->addFilter('translate', function () {
        throw new Nette\InvalidStateException('Translator has not been set.');
    });
}
```
and keep `..->addFilter` in the `Template::setTranslator` so it can overwrite the translate filter.

greeny: @matej21 Yeah, that makes more sense, implemented

src/Bridges/ApplicationLatte/TemplateFactory.php
```diff
@@ -84,6 +84,12 @@ public function createTemplate(UI\Control $control = NULL)
 			return $time == NULL ? NULL : Nette\Utils\DateTime::from($time)->modify($delta . $unit); // intentionally ==
 		});
 
+		if (!isset($latte->getFilters()['translate'])) {
+			$latte->addFilter('translate', function () {
+				throw new Nette\InvalidStateException('Translator has not been set.');
```
dg: What about advice like „… Set translator using $template->setTraslator()“

greeny: @dg added

fprochazka:
:+1:

PR #105, opened 2015-11-12
Can store your own request

Zemistr:


Zemistr:
Now you can use something like this:

```php
<?php
namespace App\Modules\BackModule\Presenters;

use Nette\Application\UI\Presenter;

class Error4xxPresenter extends Presenter {
	public function renderDefault(\Exception $exception) {
		if ($exception->getCode() === 403) {
			$this->redirect('Sign:in', ['backlink' => $this->storeRequest('+ 30 minutes', $this->getParameter('request'))]);
		}

		// load template 403.latte or 404.latte or ... 4xx.latte
		$file = __DIR__ . "/templates/Error/{$exception->getCode()}.latte";
		$this->template->setFile(is_file($file) ? $file : __DIR__ . '/templates/Error/4xx.latte');
	}
}
```

enumag:
This probably doesn't work for exceptions from createInitialRequest. I mean url `/admin/nonexistentpresenter` will trigger the global ErrorPresenter, not the one from AdminModule, right?

Zemistr:
On row 172 - 174 is fallback ;)

enumag:
Yes, that's what I'm talking about. I'd want the module specific error presenter to be triggered but instead the fallback will be used.

dg:
ad 403: this indicates one of two conditions:

- Authentication was provided, but the authenticated user is not permitted to perform the requested operation.
- The operation is forbidden to all users. For example, requests for a directory listing return code 403 when directory listing has been disabled.

Zemistr:
This change is doing simple thing "If exists error presenter in the same module, run him. If not, run global error presenter."

Zemistr:
The example is just example @dg ;)

enumag:
Yes, except that if a presenter does not exist but it is possible to determine module from routing it will be ignored.

Zemistr:
Exists another solution for this thing?
I mean, can I set different error presenter for backend and for frontend?

enumag:
@dg That's why you should extend ForbiddenRequestException and use instanceof to determine what to do with the exception instead of just error code. As an example this is sufficient.

enumag:
@Zemistr You can look here for my solution but it's not good enough to put it to nette. https://github.com/enumag/Application/blob/master/src/Application.php

Zemistr:
It seems unnecessarily complicated.
Why does my solution is not good?
I think that a modern application can not do without modules.

enumag:
I know it's complicated, but it solves the problem I mentioned above.

Zemistr:
I am little confused...what is wrong on my solution?
It works well with modules and works the same as is now in the stable release...

dg:
It seems like unexpected convention. IMHO it can be done directly in ErrorPresenter, without need to modify Application.

Zemistr:
Hmm...ok...and the second commit? :)

enumag:
@Zemistr Your solution is good except for one case. Look at the createInitialRequest method. If the presenter is not found, an exception is [thrown](https://github.com/Zemistr/application/blob/master/src/Application/Application.php#L120) and the `$request` variable which holds the information about module is lost. In this case your solution will just use global error presenter (not because of your fallback but because `$this->presenter` is null and it falls to [this](https://github.com/Zemistr/application/blob/master/src/Application/Application.php#L179)). My solution preserves the request and uses it later to trigger the correct error presenter. This is why my solution is so complicated.

Zemistr:
Okay, ignore the first commit. :)
What about the second commit?

Zemistr:
Btw.:
This is my current global error presenter :)

```php
<?php
namespace App\Modules\Presenters;

use Nette;
use Nette\Application\Responses;
use Tracy\ILogger;

class ErrorPresenter extends Nette\Object implements Nette\Application\IPresenter {
	/** @var ILogger */
	private $logger;

	public function __construct(ILogger $logger) {
		$this->logger = $logger;
	}

	public function run(Nette\Application\Request $request) {
		$exception = $request->getParameter('exception');

		if ($exception instanceof Nette\Application\BadRequestException) {
			$presenter = explode(':', $request->getParameter('request')->getPresenterName(), 2);
			$presenter[1] = 'Error4xx';

			return new Responses\ForwardResponse($request->setPresenterName(implode(':', $presenter)));
		}

		$this->logger->log($exception, ILogger::EXCEPTION);

		return new Responses\CallbackResponse(
			function () {
				require __DIR__ . '/templates/Error/500.phtml';
			}
		);
	}
}

```

enumag:
The second commit is good, I use it as well. I just have the parameters [swapped](https://github.com/enumag/Application/blob/master/src/UI/RequestStoragePresenterTrait.php#L39) with some additional magic so that I can call `$this->storeRequest($request);` without explicitly specifying expiration.

TomasVotruba:
What needs to be done here to accepted or closed?

src/Application/UI/Presenter.php
```diff
@@ -1082,17 +1082,18 @@ protected function handleInvalidLink(InvalidLinkException $e)
 
 	/**
 	 * Stores current request to session.
-	 * @param  mixed  optional expiration time
+	 * @param  mixed $expiration optional expiration time
+	 * @param  Application\Request $request optional specific request
 	 * @return string key
 	 */
-	public function storeRequest($expiration = '+ 10 minutes')
+	public function storeRequest($expiration = '+ 10 minutes', Application\Request $request = null)
```
dg: should be `NULL`

enumag:
As mentioned I'd prefer if the parameters were swapped. What do you think @dg?

src/Application/UI/Presenter.php
```diff
@@ -1082,17 +1082,18 @@ protected function handleInvalidLink(InvalidLinkException $e)
 
 	/**
 	 * Stores current request to session.
-	 * @param  mixed  optional expiration time
+	 * @param  mixed $expiration optional expiration time
+	 * @param  Application\Request $request optional specific request
```
dg: please remove `Application\Request $request`

src/Application/UI/Presenter.php
```diff
@@ -1082,17 +1082,18 @@ protected function handleInvalidLink(InvalidLinkException $e)
 
 	/**
 	 * Stores current request to session.
-	 * @param  mixed  optional expiration time
+	 * @param  mixed $expiration optional expiration time
```
dg: please remove `$expiration`

PR #106, opened 2015-12-12
SimpleRouter: add support for $defaultFlags

voda:
My use case is as described here: https://forum.nette.org/cs/3184-https-mi-automaticky-presmeruje-na-http#p93043

Its a private project, so SimpleRouter is good enough and it is running https on server and http on local machine.

### DIFF
--- a/src/Application/Routers/SimpleRouter.php
+++ b/src/Application/Routers/SimpleRouter.php
@@ -19,6 +19,9 @@ class SimpleRouter extends Nette\Object implements Application\IRouter
 	const PRESENTER_KEY = 'presenter';
 	const MODULE_KEY = 'module';
 
+	/** @var int */
+	public static $defaultFlags = 0;
+
 	/** @var string */
 	private $module = '';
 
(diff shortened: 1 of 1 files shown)


JanTvrdik:
Why not just pass `$flags` argument to constructor? Do you have multiple instances of `SimpleRouter`?

dg:
Static variables are bad…

voda:
I have router defined as a service in config.neon, so cant use a parameter because parameter needs to be set before container is created, but to detect http/https I need HttpRequest from container.

So best variant is custom factory?

enumag:
Actually you can set constructor parameters in config.neon.

https://doc.nette.org/en/2.3/configuring#toc-custom-services

voda:
I know that, but how to write `$container->getService('httpRequest')->isSecured() ? Route::SECURED : 0` in neon?

enumag:
In my opinion this flag should not be based on the current request. Why do you do that?

voda:
Because i use https on server and http on localhost. Redirect from http to https is handled by webserver.

enumag:
I solve that differently - parameter in config.neon that defaults to `Route::SECURED` but I override it to 0 in my local.neon on localhost. No need to complicate the application.

milo:
@voda 
```php
class DefaultRouterFlags
{
    public static function get()
    {
        return ....;
    }
}
```

```
services:
    - SimpleRouter(flags: DefaultRouterFlags::get())
```

voda:
Ok, thanks for the suggestions.

PR #107, opened 2015-12-15
Fix Presenter::argsToParams to not remove false

enumag:
I sometimes use boolean parameters in presenters and I want the `param=0` part in the URL because `FALSE` and `NULL` are totally different values for me. Right now however argsToParams replaces false values with nulls which breaks my use case.

### DIFF
--- a/src/Application/UI/Presenter.php
+++ b/src/Application/UI/Presenter.php
@@ -1047,7 +1047,7 @@ abstract class Presenter extends Control implements Application\IPresenter
 				));
 			}
 
-			if ($args[$name] === $def || ($def === NULL && is_scalar($args[$name]) && (string) $args[$name] === '')) {
+			if ($args[$name] === $def || ($def === NULL && $args[$name] === '')) {
 				$args[$name] = NULL; // value transmit is unnecessary
 			}
 		}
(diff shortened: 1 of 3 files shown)


matej21:
:+1:

src/Application/UI/Presenter.php
```diff
@@ -1047,7 +1047,7 @@ public static function argsToParams($class, $method, & $args, $supplemental = []
 				));
 			}
 
-			if ($args[$name] === $def || ($def === NULL && is_scalar($args[$name]) && (string) $args[$name] === '')) {
+			if ($args[$name] === $def || ($def === NULL && is_scalar($args[$name]) && !is_bool($args[$name]) && (string) $args[$name] === '')) {
```
matej21: I think, `is_scalar($args[$name]) && !is_bool($args[$name]) && (string) $args[$name] === ''` can be simplified to `$args[$name] === ''` :)

enumag: Probably. Integers and floats should never become an empty string after type casting.

enumag:
@dg Thanks!

dg:
thanks to you ;)

dg:
The question is whether to include it in v2.3…

enumag:
It admitedly is a BC break so as much as I'd like it in stable 2.x, putting it in 2.3 is probably not a best idea.

Are there other commits like this? If so we might want to release nette/application 2.4.

dg:
I did not realize that it is a BC break, so it will be in 2.4.

src/Application/UI/Presenter.php
```diff
@@ -1047,7 +1047,7 @@ public static function argsToParams($class, $method, & $args, $supplemental = []
 				));
 			}
 
-			if ($args[$name] === $def || ($def === NULL && is_scalar($args[$name]) && (string) $args[$name] === '')) {
```
Majkl578: Just wondering, why did you remove is_scalar here?

enumag:
Well since it required a change in tests and generates different URLs in some cases I would consider it a BC break...

src/Application/UI/Presenter.php
enumag: Because as far as I can tell `is_scalar($args[$name]) && (string) $args[$name] === ''` is equivalent to `$args[$name] === '' || $args[$name] === false`.

PR #108, opened 2016-01-15
UIRuntime: unnecessary required parameters

vitkutny:


src/Bridges/ApplicationLatte/UIRuntime.php
```diff
@@ -18,11 +18,11 @@
 class UIRuntime extends Nette\Object
 {
 
-	public static function renderSnippets(UI\Control $control, \stdClass $local, array $params)
+	public static function renderSnippets(UI\Control $control, \stdClass $local = NULL, array $params = [])
 	{
 		$control->snippetMode = FALSE;
 		$payload = $control->getPresenter()->getPayload();
-		if (isset($local->blocks)) {
+		if ($local && isset($local->blocks)) {
```
Majkl578: This is actually not needed, isset covers it already: https://3v4l.org/EdJa5

vitkutny: Thank you, updated

dg:
Why?

vitkutny:
I know it's marked as `@internal`. Regardless I'm using it for custom rendering. After all implementation of that method really does not require those params and it won't affect anyone else, I think.
snimek obrazovky 2016-01-15 v 9 50 04

PR #111, opened 2016-01-19
json response: removed automatic cache disabling

hrach:
The json output isn't something special what shouldn't be cached at all. If disabling of cache is needed, this should be handled by Nette Presenter itself.

It's probably BC Break.

### DIFF
--- a/src/Application/Responses/JsonResponse.php
+++ b/src/Application/Responses/JsonResponse.php
@@ -62,7 +62,6 @@ class JsonResponse extends Nette\Object implements Nette\Application\IResponse
 	public function send(Nette\Http\IRequest $httpRequest, Nette\Http\IResponse $httpResponse)
 	{
 		$httpResponse->setContentType($this->contentType);
-		$httpResponse->setExpiration(FALSE);
 		echo Nette\Utils\Json::encode($this->payload);
 	}
 


Majkl578:
:+1: Disabling caching doesn't make sense.

fprochazka:
Is there a way to override this? Or you just couldn't use the response object if you wanned it to be cached? 

Maybe an option would be better than removing it completely.

enumag:
:+1: Makes sence.

dg:
Can you change `json response` to `JsonResponse` in commit message?

hrach:
Repushed with BC break info.

enumag:
Maybe we should add `$expiration` parameter to `Presenter::sendJson();`? Not sure about what the default value should be though.

Majkl578:
Why? You can configure it directly through Http\Response object.

hrach:
I don't see the added complexity of another parametr. Feel free to change change the expiration in the Response instance.

enumag:
Well to achieve the original behavior I'll have to replace all of my occurences of sendJson with

```php
$response = new JsonResponse($data);
$response->setExpiration(false);
$this->sendResponse($response);
```

The parameter would simplify it pretty well.

Majkl578:
Or you can just use:
```php
$this->getResponse()->setExpiration(FALSE);
$this->sendJson(...);
```

hrach:
Thanks!

PR #112, opened 2016-01-21
Typehinted integer in PHP 7 is ommited from generated URL when it's zero

ondrejmirtes:
The attached commit shows the described situation.

Expected result of `$this->link('edit', ['id' => 0])`:

> /index.php?id=0&action=edit&presenter=Test

Actual:

> /index.php?action=edit&presenter=Test


ondrejmirtes:
I would love to fix it but honestly don't know how :( PresenterComponentReflection::convertType() correctly process the parameter as zero.

dg:
Actual result is IMHO fine.

hrach:
IMO is't a bug.

ondrejmirtes:
I think it's a bug too. "int" represents all integers including negative ones and zero and I don't see a reason why a `0` integer should be treated differently.

dg:
This `?id=0` can be achieved in this way:

`public function actionEdit(int $id = NULL)`

ondrejmirtes:
`actionEdit(int $id = NULL)` means something different to me, it means that the parameter `$id` is optional and nullable. But `actionEdit(int $id)` means that `$id` is required and ranges from PHP_INT_MIN to PHP_INT_MAX and it cannot be `NULL`.

Majkl578:
I too think it's a bug unless the action is with default value (`actionFoo(int $id = 0)`). 0 is a valid value.

dg:
So what it should do when `a=0` is not in URL?

ondrejmirtes:
> So what it should do when a=0 is not in URL?

This question is about matching URLs and this PR is about constructing them.

But I'd say in this case the result of `$this->getParameter('id')` is null and null cannot be passed to `int $id`, so it's basically a type mismatch.

hrach:
When a is not present, null should be passed to method call, an php exception will be triggered and handled by the app's error presenter.

dg:
i.e. it should throw 500 error? Ohh…

ondrejmirtes:
It means that the router is misconfigured/implemented poorly so I'm fine
with 500 too.

On Thursday, 21 January 2016, David Grudl <notifications@github.com> wrote:

> i.e. it should throw 500 error? Ohh…
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nette/application/pull/112#issuecomment-173564355>.
>


-- 

Ondřej Mirtes

Majkl578:
> It means that the router is misconfigured/implemented poorly so I'm fine with 500 too.

:+1:

dg:
It is not about router. We are talking about `?a=0` in query.

And 500 means server error. Server error due to the removal of a single parameter from URL? Boys! Do not be silly :-)

JanTvrdik:
As I proposed couple of years ago, missing required parameter should result in BadRequestException#404.

ondrejmirtes:
I want the HTTP request to give me an integer. If there's no integer in the request, I have to respond with an error. What kind of error is a subject to discuss.

I think this layer of request processing should be stopped at the router level. If an incorrect type is received by a presenter (involving some lossless type casting), something is very wrong and I would like this situation to result in a new error log line. So I'm voting for a 500 error.

But if Nette wants to be more loose in these situations, I can accept throwing 404 too.

Majkl578:
@dg: The issue was about generated URLs, not their handling.
If you require a parameter in the action and it happens to be missing in the request, it is obviously not OK. What @JanTvrdik proposes makes sense in such case.

Also I see a huge difference now that we have PHP 7 with scalar types. In the old times we often used default value (e.g.  `actionFoo($id = 0)`) just to express the default type, but it wasn't a invalid signature as the parameter is required. With PHP 7 we no longer need to use this _hack_ as we can use proper typehints.

JanTvrdik:
Possibly related (5 years old) forum thread https://forum.nette.org/cs/6819-validace-parametru-v-presenterech but due to it's age lot's of information and ideas may be outdated

hrach:
> i.e. it should throw 500 error? Ohh…

It doesn't mean, that the resource does not exist. You didn't fullfill the contract, server failed to do his job. That's all.

I'm probably ok with 400. But 404 is non-sence.

pilec:
Btw, you probably look for 422: http://www.restpatterns.org/HTTP_Status_Codes/422_-_Unprocessable_Entity

ondrejmirtes:
Can we go back to the original issue of constructing URL with zero integer as a parameter and create a new thread for support for scalar typehints in presenters and how to solve receiving requests with them? Thank you.

dg:
> @JanTvrdik As I proposed couple of years ago, missing required parameter should result in BadRequestException#404

This is obviously a good idea, but the giant BC break.
But it would be possible with scalar typehints because they are new (so no BC break), just inconsistency.

> @ondrejmirtes I want the HTTP request to give me an integer.

It does :-) It gives you 0. Whether it is, number of page or id, you have to validate it.

> If there's no integer in the request, I have to respond with an error. What kind of error is a subject to discuss.

Absolutely not. There's nothing to discuss, it just needs to be BadRequestException. (ie. 4xx, imho 404 is fine)

> I think this layer of request processing should be stopped at the router level. 

How can you define in router, that the query must contains a parameter `a`? And numeric.

@Majkl578 The issue was about generated URLs, not their handling.

Generating and matching are very close. We cannot discuss about first one and aviod discussion about second one.

ondrejmirtes:
If there's a missing parameter in the request, I don't want a **random** integer. Zero has the same weight as any other integer.

> How can you define in router, that the query must contains a parameter a? And numeric.

`/admin/product/<presenter>/<action>[/<id [1-9][0-9]*>]` - this is how we define positive integers in `id` parameter in masks in `Nette\Application\Routers\Route`.

dg:
> @ondrejmirtes  Can we go back to the original issue of constructing URL with zero integer as a parameter and create a new thread for support for scalar typehints in presenters and how to solve receiving requests with them? Thank you.

No, it is complex problem. You have to take into account the compatibility. What happens when someone will upgrade PHP 5 `renderDetail($id = 0)` to PHP 7 `renderDetail(int $id)`. In old links there is no `id=0`. Etc.

ondrejmirtes:
`renderDetail($id = 0)` and `renderDetail(int $id)` mean something different, the first one has default value and the second one does not. Correct refactoring to PHP 7 code is from `renderDetail($id = 0)` to `renderDetail(int $id = 0)`.

hrach:
> Correct refactoring to PHP 7 code is from renderDetail($id = 0) to renderDetail(int $id = 0).

:+1:

Majkl578:
@pilec: 422 is not valid for this case, because request URI is not an entity (entity = http body).

dg:
1) we can responds with 404 when parameter with scalar type and without default value is missing
- no BC break *
- solves this issue
- makes inconsistency with typehint `array` and with parameters without typehint

2) we can responds with 404 when any parameter withnout default value is missing
- giant BC break
- solves this issue
- all parameters are handler the same way

3) we can pass default values to parameters with scalar type and without default value (current behavior)
- no BC break
- solving this issue will create inconsistency how Nette generates URL
- all parameters are handler the same way

I think that 2) is impossible and 1) vs 3) is about consistency (ie. a about different handling of `array $arr` vs. `int $id` etc). 

IMHO 1) is best way.

hrach:
> IMHO 1) is best way.

:+1: , still, I'm more for 400.

dg:
400 is perfectly ok, I'll try to find out how it affects Google. For example, 410 is big problem, because Google will ignore this URL, although it becomes exist in future.

ondrejmirtes:
400 is good from the point of view of the HTTP spec, but it will likely cause people having to write a new latte template or change their code in the ErrorPresenter.

JanTvrdik:
I think that the best choice for Nette 3 is (1) + change array typehint behavior to match other typehints (BC break) + possibly implement support for class typehints (e.g ˙show (Article $article)˙)

dg:
Class type hints are IMHO not needed, because object is created in router, which is coupled with presenter.

dg:
I forgot to mention persistent parameters - so 1) makes inconsistency between typehinted parameters and persistent parameters, because properties in PHP have no typehints. It is possible to use annotation `@var`, but it will be BC break.

dg:
> @JanTvrdik change array typehint behavior to match other typehints (BC break) 

It is tricky, because it is not possible to have empty array in URL.

JanTvrdik:
Random thoughts:

* We could make `array $list` mean "at least 1 item" and make `array $list = []` mean "any number of items".
* You cannot have empty array in URL but you can have empty array in application request.

ondrejmirtes:
Cool, thanks! :+1:

hrach:
Great! Thank you!

fprochazka:
Looks great :+1:

Issue #113, opened 2016-02-01
New interface for Route display in debugbar


h4kuna:
If I want create new Route and display values in debug bar, i must extend Route class. What do you think about new interface with two methods?
```php
interface IRouteBar {
public funciton getDefaults();
public function getMask();
}
```
Class route must refaktor.
```php
class Route implements IRoute, IRouteTracy {}
```

And I will use like this
```php
class MyRoute implements IRoute, IRouteTracy {}
```
instance of
```php
class MyRoute extends Route {}
```

Or append both methods to IRoute interface?

fprochazka:
@h4kuna interface is a good idea, just a suggestion: `IRouteBar` -> `IRouteMeta`

PR #114, opened 2016-02-01
Interface IRouteMeta define API for RoutingPanel.

h4kuna:
Issue #113


src/Application/IRouteMeta.php
```diff
@@ -0,0 +1,29 @@
...
+ * The bi-directional router.
+ */
+interface IRouteMeta
+{
+
+	/**
+	 * Return route mask.
+	 * @return string
+	 */
+	public function getMask();
```
Majkl578: Nette doesn't use visibility in interfaces.

Majkl578:
I don't like the name, it sounds like an abbreviation. :/

src/Application/IRouteMeta.php
h4kuna: Ok, fixed.

h4kuna:
@Majkl578 What do you suggest?

I named in issue like IRouteBar? What do you think about IRouteInfo, IRouteMask, IRouteMaskBar?

JanTvrdik:
Generally good idea but the issue is much more complex.

h4kuna:
@JanTvrdik Let's describe it.

JanTvrdik:
You need to consider custom routers (which can do pretty much anything) as well. Some custom routers may be just wrappers of another routers, some do complex logic.

The entire interface of routing bar may need to be redesigned to be able to clearly display advanced router configurations.

foxycode:
@JanTvrdik You are true, but I think this is good step forward.

PR #115, opened 2016-02-03
ApplicationExtension: do not initialize RoutingPanel if Tracy is not installed

xificurk:
A bit smarter debugging initialization in ApplicationExtension.


dg:
Thanks

PR #116, opened 2016-02-08
TemplateFactory: removed useless code

xificurk:
Can't figure out why this was added in the first place, all tests are passing... I guess it's useless.


dg:
This is really unnecessary code, the plan was to move finding of templates from Presenter to this loader. It can be removed in 2.3 (but I am not sure whether it is BC or not), but it can be finished in master.

xificurk:
I see... thanks for clarification.

The only way I can see, that this could break things is if someone is setting custom `Loader` in `LatteFactory` service. Current `TemplateFactory` code essentially always forces `FileLoader` instead of whatever `LatteFactory` created.
In that case, I might even consider this PR  a bugfix ;-)

PR #118, opened 2016-02-10
Support for direct setting layout in presenter

besir:


src/Application/UI/Presenter.php
```diff
@@ -475,7 +475,10 @@ public function findLayoutTemplateFile()
 	{
 		if ($this->layout === FALSE) {
 			return;
+		} elseif (file_exists($this->layout)) {
```
fprochazka: I think having some check for `/` or `\`, before the IO call would be faster, than always doing a `file_exists`.

fprochazka:
It's not a bad idea, although I never needed it myself. A test case would be nice :)

besir:
Ok, I will check first letter of the string and after check file_exists(), I dont have Idea for test :-/ THis method have not any test. Can U help me @fprochazka? THX

besir:
```
if ($this->layout === FALSE) {
			return;
		} elseif (in_array(substr($this->layout, 0, 1), ['/', '\\']) && file_exists($this->layout)) {
			return $this->layout;
		}
```
??

dg:
` } elseif (preg_match('#[/\\\\]#', $this->layout) && file_exists($this->layout)) {`

besir:
Why Regexp? Only speed or have any next positives in this context? THX David.

http://www.knesl.com/articles/view/nepouzivejte-regularni-vyrazy

dg:
It checks whether string contains \, not only in first character.

besir:
Its fact, file path dont must be absolute path... Ok, I will test it and i make test (HUH) :D . Thx for cooperatinon.

dg:
Merged e2c1a1da7e3451744c70593f8f38c937cb570a3b

Issue #120, opened 2016-02-19
Set charset for JsonResponse


patrickkusebauch:
When sending a JsonResponse, you cannot set the JSON charset right now. There is a work around right now with
`$this->sendResponse(new \Nette\Application\Responses\JsonResponse($data, 'application/json; charset=utf-8'));
`, but it ?should not? work.

There is a [discussion](https://forum.nette.org/cs/25615-cestina-v-jsonu-se-nezobrazuje-korektne) of this behavior on Nette Framework Forum (in Czech).

I am willing to write a PR, of somebody decides the solution (you should not be able to set charset under any circumstances vs add ability to set charset transparently).

dg:
Nette supports only UTF-8.

dg:
Are you saying that there is missing charset=UTF-8? So please send PR and set 'utf-8' here https://github.com/nette/application/blob/a756b711ec741b6db58c8590a6251b9c78c9fffa/src/Application/Responses/JsonResponse.php#L64, `setContentType($this->contentType, 'utf-8')`

patrickkusebauch:
Yes, that is exactly what I am saying. I will make a PR over the weekend.

JanTvrdik:
This is more of a Chrome bug than Nette bug because 
1) encoding of JSON can be reliable determined from first few bytes
2) only UTF-8 is used in real-world

DrunkedBrigand:
Try this:
$this->payload = ["test"=>"ÝŽáýžřáýž"];
$this->sendPayload();

and you will see Chrome bug.
In real world are used ALL possible encodings. Even the impossible ones.

patrickkusebauch:
@dg Furthermore, there is a possible bug in Nette\Http\Repsonse, where you can set the charset by calling` Nette\Http\Response::setContentType('application/json; charset=utf-8', null) `instead of `Nette\Http\Response::setContentType('application/json', 'charset=utf-8')`

dg:
https://bugs.chromium.org/p/chromium/issues/detail?id=438464

jahudka:
Well this is just wrong :-( There *is* no "charset" parameter for the application/json content-type *and* JSON *must* be unicode by definition. Setting a charset to an application/json response may break things, like Nittro for example, because some people are trying to follow standards and specs and those specifically say there *is no charset for application/json*...

jahudka:
Besides, if you read through the chromium bug report, this is an edge case at best..
[...]

PR #121, opened 2016-02-20
Set default JsonResponse charset to utf-8.

patrickkusebauch:
- bugfix
- issues #120 
- documentation - not needed
- BC break - yes (removes unintended feature see [Nette Forum discussion (in Czech)](https://forum.nette.org/cs/25615-cestina-v-jsonu-se-nezobrazuje-korektne))


patrickkusebauch:
If I understand correctly, it seems lie "length" is now a default filter in Latte. (with [this commit](https://github.com/nette/latte/commit/5d5520a78b279a9571698fe8f46a009abeaf16cf) on Feb 19 2016)

patrickkusebauch:
This PR is fixed by PR #122.

dg:
Thanks, merged 2f939c8ddcf877af8f0cd3eee240489587edbc09

PR #125, opened 2016-03-28
Tests for isLinkCurrent

ondrejmirtes:

Issue #124, opened 2016-03-21
2.3.11 breaks isLinkCurrent and {ifCurrent} macro


ondrejmirtes:
Following-up on my [bug report](https://github.com/nette/application/pull/112) about typehinted scalar parameters. The fix that resolved this bug also introduced a rather inconvenient BC break.

With non-typehinted scalar parameters, it's not necessary to pass all parameters to isLinkCurrent method:

`$presenter->isLinkCurrent(':Homepage:default', []);`

In this example, calling the method works even if the requested action has required arguments. By calling it without the arguments, I simply mean to ask "Is the current page at Homepage:default?" without checking exact parameters.

If the action has PHP-7-scalar typehints, I cannot longer call isLinkCurrent without its arguments:

```php
public function renderDefault(int $id)
{
}
...
$presenter->isLinkCurrent(':Homepage:default', []); // fails on PHP-7
```

BC break aside, I think it also reduces the scope where isLinkCurrent is usable. Before PHP 7, I could use it for checking whether am I in a concrete Presenter:action pair without caring about exact arguments.

For example if I want to always render a template block when I am at `Article:default`, in the past I could have used `{ifCurrent Article:default}`. Now, I have to pass the arguments to the macro.

Ideally, the isLinkCurrent/ifCurrent functionality should not depend on generating links because it should behave independently of each other because of different requirements for these functionalities.

If this issue is a logical conclusion of the suggested behaviour in #112 and the current behaviour is fine for everyone, feel free to close this issue.

Thanks!

dg:
Good point, can you send PR?

ondrejmirtes:
I opened #125 which tests current behaviour of isLinkCurrent - I found out there are no tests for this method so I added several test cases and refactored tests to be more compliant with DRY.

I'd like if you'd verified that the tests are OK because they will be a foundation for the isLinkCurrent rewrite.

I also opened #126 which now contains failing tests for PHP 7. The asserts for isLinkCurrent are now reused across PHP < 7 and >= 7 versions so the difference in behaviour is quite clear.

ondrejmirtes:
#125 is intended to be merged into master before #126 because it just tests the current functionality. There's nothing blocking that once we are satisfied with the tests.

PR #126, opened 2016-03-28
isLinkCurrent compatible with PHP 7 typehints

ondrejmirtes:


ondrejmirtes:
@dg This is a first draft aimed at resolving #124. I added a bunch more tests that should have been part of PR #125, but I didn't foresee the need for them.

I wasn't sure what data structure should I use to pass data from parseLinkDestination, value object with getters might be better but I'm not sure about performance impact so I sticked to a plain array.

src/Application/UI/Presenter.php
```diff
@@ -779,87 +779,23 @@ protected function createRequest($component, $destination, array $args, $mode)
...
+		$parsedLinkDestination = $this->parseLinkDestination($component, $destination);
+		$destination = $parsedLinkDestination['destination'];
+		$scheme = $parsedLinkDestination['scheme'];
+		$presenter = $parsedLinkDestination['presenter'];
+		$presenterClass = $parsedLinkDestination['presenterClass'];
+		$action = $parsedLinkDestination['action'];
+		$current = $parsedLinkDestination['current'];
+		$component = $parsedLinkDestination['callee'];
+		$signal = $parsedLinkDestination['signal'];
+		$fragment = $parsedLinkDestination['fragment'];
```
JanTvrdik: You can use `extract`

ondrejmirtes: That's exactly what I didn't want to do.

Ondřej Mirtes




On Tue, Mar 29, 2016 at 2:35 AM -0700, "Jan Tvrdík" <notifications@github.com> wrote:












In src/Application/UI/Presenter.php:
> -			}
> -			try {
> -				$presenterClass = $this->presenterFactory->getPresenterClass($presenter);
> -			} catch (Application\InvalidPresenterException $e) {
> -				throw new InvalidLinkException($e->getMessage(), NULL, $e);
> -			}
> +		$parsedLinkDestination = $this->parseLinkDestination($component, $destination);
> +		$destination = $parsedLinkDestination['destination'];
> +		$scheme = $parsedLinkDestination['scheme'];
> +		$presenter = $parsedLinkDestination['presenter'];
> +		$presenterClass = $parsedLinkDestination['presenterClass'];
> +		$action = $parsedLinkDestination['action'];
> +		$current = $parsedLinkDestination['current'];
> +		$component = $parsedLinkDestination['callee'];
> +		$signal = $parsedLinkDestination['signal'];
> +		$fragment = $parsedLinkDestination['fragment'];



You can use extract

dg: $parsedLinkDestination['current'] is always FALSE

dg:
There is something wrong, now test passes even if you delete all methods from TestControl…

dg:
`tests: Moved TestControl to separate file to be reusable across tests` I think it is not good idea, it brings dependency between different tests. I'll skip this commit and merge other tests to master, ok?

dg:
@ondrejmirtes I closed it this way 6d12b3c6330896fe0ec0cf0d3dd2956a5636b10b, and thanks for all the tests!

ondrejmirtes:
Cool :+1:

PR #130, opened 2016-04-08
Fix wrong MicroPresenter module

grongor:
Before this change I was getting error `Case mismatch on presenter name 'Nette:Micro', correct name is 'NetteModule:Micro'.`

I hope this is correct change - it is the first time I'm using the MicroPresenter. 

My usage was as simple as:
```
$this[] = new Route('/', function() {
    // some simple stuff
});
```


dg:
Did you change presenter mapping?

dg:
ping @grongor

grongor:
Wow, really sorry, I totally forgot about that one. Yeah, of course, I did change the mapping and I did not understand "how it works" so I changed it wrong :) Sorry for taking your time ...

PR #132, opened 2016-05-01
Remove dependency on nette/security

enumag:
As far as I can tell nette/security seems to be optional everywhere in nette/application.

### DIFF
--- a/src/Application/UI/Presenter.php
+++ b/src/Application/UI/Presenter.php
@@ -1080,7 +1080,7 @@ abstract class Presenter extends Control implements Application\IPresenter
 			$key = Nette\Utils\Random::generate(5);
 		} while (isset($session[$key]));
 
-		$session[$key] = [$this->getUser()->getId(), $this->request];
+		$session[$key] = [$this->user ? $this->user->getId() : null, $this->request];
 		$session->setExpiration($expiration, $key);
 		return $key;
 	}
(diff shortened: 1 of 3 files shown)


dg:
Can you change commit message to smthng like `composer: removed …`

enumag:
@dg Done.

dg:
Btw it is required by storeRequest() & restoreRequest()

enumag:
@dg Added some conditions to make them work without user.

dg:
Did you test it?

enumag:
Oh right... can't use getUser because it throws an exception... Added a test for storeRequest as well.

Can't really test it in a real application though. I use my own implementation of these methods to make them work with object parameters.

Majkl578:
Needs rebase after force-pushes... :(

dg:
Thanks, merged

src/Application/UI/Presenter.php
```diff
@@ -1080,7 +1080,7 @@ public function storeRequest($expiration = '+ 10 minutes')
 			$key = Nette\Utils\Random::generate(5);
 		} while (isset($session[$key]));
 
-		$session[$key] = [$this->getUser()->getId(), $this->request];
+		$session[$key] = [$this->user ? $this->user->getId() : null, $this->request];
```
lookyman: `NULL`

enumag: Yeah, @dg already fixed that when he merged it. :-)

PR #134, opened 2016-05-04
PresenterComponent ⇒ Component and PresenterComponentReflection ⇒ ComponentReflection

dg:


src/Application/UI/Component.php
```diff
@@ -20,7 +20,7 @@
  * @property-read Presenter $presenter
  * @property-read bool $linkCurrent
  */
-abstract class PresenterComponent extends Nette\ComponentModel\Container implements ISignalReceiver, IStatePersistent, \ArrayAccess
+abstract class Component extends Nette\ComponentModel\Container implements ISignalReceiver, IStatePersistent, \ArrayAccess
```
fprochazka: I've seen people extend this class in real world.

dg: You right. This is just WIP, so alias is missing. Probably class_alias should be used.

PR #135, opened 2016-05-05
VaribaleTemplate: add getParameters() method to improve back compatability

xificurk:
Quite a few custom macros use something like: https://github.com/brabijan/images/blob/master/src/Brabijan/Images/Macros/Latte.php#L138

### DIFF
--- a/src/Bridges/ApplicationLatte/VariableTemplate.php
+++ b/src/Bridges/ApplicationLatte/VariableTemplate.php
@@ -78,4 +78,14 @@ class VariableTemplate extends Latte\Template
 		unset($this->template->params[$name]);
 	}
 
+
+	/**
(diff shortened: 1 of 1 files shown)


xificurk:
Hm, travis failed, but as far as I can tell, it's unrelated with this change.

dg:
Hmmm, there is not getParameters()? It extends Latte\Template.

xificurk:
There is, but it's returning parameters of `VariableTemplate` (which is empty array). You need to return parameters of the underlying latte template object.

dg:
Oh! Please can you fix typo in commit message: `VariableTemplate`

xificurk:
Good catch! Fixed.

dg:
Thanks

PR #137, opened 2016-05-06
Latte bridge: fixed missing $template variable in a snippet

matej21:


dg:
I do not get how it is possible that the parent template has not `$template` variable…

matej21:
The variable is not available in the child template, not the parent. That is because `render` prolog looks like this:
```php
// prolog Latte\Macros\CoreMacros
if ($this->initialize($_args)) return;
extract($_args);	
// prolog Nette\Bridges\ApplicationLatte\UIMacros
if (Nette\Bridges\ApplicationLatte\UIRuntime::initialize($this, $this->blockQueue)) return;
$template = $this->params["template"];
```

so when a template has a layout, then render method is terminated (by `$this->initialize()`) before UIRuntime::initialize is called.

dg:
What about to add `$template->template = $template;` here https://github.com/nette/application/blob/master/src/Bridges/ApplicationLatte/TemplateFactory.php#L96 and remove whole VariableTemplate?

PR #139, opened 2016-06-02
SnippetBridge: Remove IRenderable check (Control is always IRenderable)

Majkl578:
Root control is always IRenderable since it can only be Control (which implements IRenderable).
cc @matej21


dg:
Thanks

PR #142, opened 2016-06-10
LinkGenerator: throws exception, when try to pass non-associative parameters to missing action

davidkrmela:
This fixes inconsistent between `$presenter->link` and `$linkGenerator->link`. Presenter throws exception, but LinkGenerator returns URL, for example `http://www.example.com/presenter/action?0=10`.

### DIFF
--- a/src/Application/LinkGenerator.php
+++ b/src/Application/LinkGenerator.php
@@ -62,6 +62,8 @@ class LinkGenerator
 				|| method_exists($class, $method = $class::formatRenderMethod($action))
 			) {
 				UI\Presenter::argsToParams($class, $method, $params);
+			} elseif (array_key_exists(0, $params)) {
+				throw new UI\InvalidLinkException("Unable to pass parameters to action '$presenter:$action', missing corresponding method.");
 			}
 		}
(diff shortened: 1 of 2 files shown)


dg:
:+1:

src/Application/LinkGenerator.php
```diff
@@ -62,6 +62,8 @@ public function link($dest, array $params = [])
 				|| method_exists($class, $method = $class::formatRenderMethod($action))
 			) {
 				UI\Presenter::argsToParams($class, $method, $params);
+			} elseif (array_key_exists(0, $params)) {
```
hrach: should be optimized with isset(), links generation is quite repeated task.

Majkl578: That would allow `0 => NULL`.

JanTvrdik: @hrach probably meant `isset($params[0]) || array_key_exists(0, $params)`

hrach: Yes, that was what I meant, but the or discards the added value :(

PR #143, opened 2016-06-17
Request wrapper class now contains request body.

simPod:
we have added body to the request as we use it in our rest api presenters


dg:
Why?

PR #146, opened 2016-07-21
FIX: Unescaped preg_match_all parameter

miloshavlicek:
Fixed unescaped annotation name $name in preg_match_all function $pattern parameter. (Causing errors if annotation is like @Secured\User - ipub/permissions package)

### DIFF
--- a/src/Application/UI/ComponentReflection.php
+++ b/src/Application/UI/ComponentReflection.php
@@ -195,7 +195,7 @@ class ComponentReflection extends \ReflectionClass
 	 */
 	public static function parseAnnotation(\Reflector $ref, $name)
 	{
-		if (!preg_match_all('#[\\s*]@' . preg_quote($name) . '(?:\(\\s*([^)]*)\\s*\))?#', $ref->getDocComment(), $m)) {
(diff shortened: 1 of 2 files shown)


src/Application/UI/ComponentReflection.php
```diff
@@ -195,7 +195,7 @@ public static function convertType(& $val, $type, $isClass = FALSE)
 	 */
 	public static function parseAnnotation(\Reflector $ref, $name)
 	{
-		if (!preg_match_all("#[\\s*]@$name(?:\(\\s*([^)]*)\\s*\))?#", $ref->getDocComment(), $m)) {
+		if (!preg_match_all('#[\\s*]@' . preg_quote($name) . '(?:\(\\s*([^)]*)\\s*\))?#', $ref->getDocComment(), $m)) {
 			return FALSE;
```
dg: There should be `preg_quote($name, '#')`

dg:
Thanks

src/Application/UI/ComponentReflection.php
miloshavlicek: Yes, I'm sorry... fixed in next diff

PR #148, opened 2016-07-27
Presenter: fixed creating link with required persistent parameters

JanTvrdik:


src/Application/LinkGenerator.php
```diff
@@ -61,7 +61,13 @@ public function link($dest, array $params = [])
 			if (method_exists($class, $method = $class::formatActionMethod($action))
 				|| method_exists($class, $method = $class::formatRenderMethod($action))
 			) {
-				UI\Presenter::argsToParams($class, $method, $params);
+				UI\Presenter::argsToParams($class, $method, $params, [], $missing);
+				if ($missing) {
+					$argName = key($missing);
```
dg: This can be shorted to `list($argName, $rm) = each($missing)`

dg:
:+1:

PR #149, opened 2016-07-27
Wrong argument for createRequest method

semerak-michal:
Typo - redirectX to redirect. If it was not the intention.


dg:
It is intention.

PR #151, opened 2016-08-01
ComponentReflection::parseAnnotation: fix false positive annotation parsing

mrtnzlml:
I know it's internal class, but I am using it anyway - so it's maybe my problem. And I know about this issue quite long, but [with this commit](https://github.com/nette/application/commit/03b8b34a0518b7adc09883564dccaacde738af22) I have to fix it. In current master you can have `@privatex` annotation and this test will pass:

```php
Assert::same([TRUE], Reflection::parseAnnotation($rc, 'private'));
```

But I don't think this is right. Previously I was calling it like this:

```php
$visible = ComponentReflection::parseAnnotation($actionReflection, 'public\\s');
```

But it's not possible now... :)

### DIFF
--- a/src/Application/UI/ComponentReflection.php
+++ b/src/Application/UI/ComponentReflection.php
@@ -195,7 +195,7 @@ class ComponentReflection extends \ReflectionClass
 	 */
 	public static function parseAnnotation(\Reflector $ref, $name)
 	{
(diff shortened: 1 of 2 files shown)


dg:
Good point. `\w` is not enough, char `-` must not follow too. Probably better will be white-list, i.e. `'(?=\s|\(|$)(?:\(\s*([^)]*)\s*\))?#'``

mrtnzlml:
Yes, I need to practice regular expressions more - thanks. Updated... :)

dg:
Great, thx!

PR #153, opened 2016-08-14
LinkGenerator respect HTTP_X_FORWARDED_PROTO header

Vrtak-CZ:


dg:
It smells. Hidden dependencies on global variables.

JanTvrdik:
@Vrtak-CZ I don't understand what you are trying to solve. This should be handled by `Http\RequestFactory`.

Vrtak-CZ:
Sorry guys I just misconfigure trusted proxy in config.

PR #154, opened 2016-08-15
Fix possible PHP warning "strpos(): Offset not contained in string"

enumag:
It happened in some of my failing tests because `$refUrl->getScheme()` was empty for whatever reason.

Anyway a PHP warning should be fixed.


dg:
Thanks

PR #158, opened 2016-09-22
Call checkRequirements for createComponent methods

enumag:


TomasVotruba:
Thank you for creating PR.

How do you solve this now?

enumag:
With a trait that has exactly the same implementation of createComponent.

enumag:
@dg What do you think about this for Nette 3.0?

dg:
I think it's ok.

enumag:
ping @dg

Issue #141, opened 2016-06-07
TemplateFactory - variable template class (RFC)


f3l1x:
Hi! I have and idea about TemplateFactory.

If you need to use your own Template class, you have to extend original TemplateFactory now.

What about this implementation?

``` php
protected $templateClass = 'Nette\Bridges\ApplicationLatte\Template';

public function createTemplate(UI\Control $control = NULL)
{
    $latte = $this->latteFactory->create();
    $template = new $this->templateClass($latte);
    $presenter = $control ? $control->getPresenter(FALSE) : NULL;
    //...
}

public function setTemplateClass($class) 
{
    $this->templateClass = $class;
}

```

What do you think? It makes TemplateFactory more flexible. You can setup your template implementation over NEON config. 

``` yaml
services:
    latte.templateFactory:
        setup:
            - setTemplateClass(My\Custom\Template)
```

---

Thanks for feedback.

TomasVotruba:
What is the use case for this?

mrtnzlml:
@TomasVotruba For example [this](http://zlml.cz/dva-sablonovaci-systemy-zaroven#toc-templatefactory).

TomasVotruba:
@mrtnzlml I see.

Coold you prepare PR for this with test?

mrtnzlml:
I implemented it differently. But in fact it's quite similar to the @f3l1x solution. The only difference is that my solution doesn't use setter. And even though I believe that it's better without setter, I'm not sure about my implementation. Usage:

```
services:
	latte.templateFactory:
		arguments:
			templateClassName: App\CustomTemplateImplementation
```

Unfortunately Template cannot be service (or at least I don't know how to do it properly - and I tried).

f3l1x:
Maybe it's little bit confusing name `Template Factory`, but I don't know better. What about new concept?

- LatteFactory - creates latte engine
- TemplateFactor - creates `TemplateFactory` via `LatteFactory` and `TemplateClassFactory`
- TemplateClassFactory - creates just Template class  (depends on `TemplateFactory`)

mrtnzlml:
Yeah, I was thinking about this approach but I want to know more opinions (from more people) first... :)

TomasVotruba:
Please discuss under PR, as it is suggested solution and it's more relevant there.
Thank you

PR #159, opened 2016-09-25
TemplateFactory: add support for custom Template implementation

mrtnzlml:
With this change it's possible to simply use custom Template implementation. Before this commit Template object initialization was hard-coded into TemplateFactory (related issue: https://github.com/nette/application/issues/141). Usage example (`config.neon`):

```
latte:
    templateClass: App\CustomTemplateImplementation
```

Custom template implementation MUST implement Nette\Application\UI\ITemplate interface.
- feature
- issues - #141
- documentation - probably not needed because this part of framework is not documented (internals)
- BC break - no

This is more request for comments than real PR. Thank you.

### DIFF
--- a/src/Bridges/ApplicationDI/LatteExtension.php
+++ b/src/Bridges/ApplicationDI/LatteExtension.php
@@ -19,6 +19,7 @@ class LatteExtension extends Nette\DI\CompilerExtension
 	public $defaults = [
 		'xhtml' => FALSE,
 		'macros' => [],
+		'templateClass' => NULL
 	];
 
 	/** @var bool */
@@ -52,9 +53,12 @@ class LatteExtension extends Nette\DI\CompilerExtension
 			->addSetup('Nette\Utils\Html::$xhtml = ?', [(bool) $config['xhtml']])
 			->setImplement(Nette\Bridges\ApplicationLatte\ILatteFactory::class);
 
-		$builder->addDefinition($this->prefix('templateFactory'))
+		$templateFactory = $builder->addDefinition($this->prefix('templateFactory'))
 			->setClass(Nette\Application\UI\ITemplateFactory::class)
 			->setFactory(Nette\Bridges\ApplicationLatte\TemplateFactory::class);
+		if ($config['templateClass'] !== NULL) {
+			$templateFactory->addSetup('?->setTemplateClass(?)', ['@self', $config['templateClass']]);
+		}
 
 		foreach ($config['macros'] as $macro) {
 			if (strpos($macro, '::') === FALSE && class_exists($macro)) {
--- a/src/Bridges/ApplicationLatte/TemplateFactory.php
+++ b/src/Bridges/ApplicationLatte/TemplateFactory.php
@@ -31,6 +31,9 @@ class TemplateFactory implements UI\ITemplateFactory
 	/** @var Nette\Caching\IStorage */
 	private $cacheStorage;
 
+	/** @var string */
+	private $templateClass = Template::class;
+
 
 	public function __construct(ILatteFactory $latteFactory, Nette\Http\IRequest $httpRequest = NULL,
 		Nette\Security\User $user = NULL, Nette\Caching\IStorage $cacheStorage = NULL)
@@ -48,7 +51,7 @@ class TemplateFactory implements UI\ITemplateFactory
 	public function createTemplate(UI\Control $control = NULL)
 	{
 		$latte = $this->latteFactory->create();
-		$template = new Template($latte);
+		$template = new $this->templateClass($latte);
 		$presenter = $control ? $control->getPresenter(FALSE) : NULL;
 
 		if ($control instanceof UI\Presenter) {
@@ -110,6 +113,14 @@ class TemplateFactory implements UI\ITemplateFactory
 		}
 
 		return $template;
+	}
+
+	public function setTemplateClass($templateClass)
+	{
+		if (!class_exists($templateClass) || !is_a($templateClass, Template::class, TRUE)) {
+			throw new Nette\InvalidArgumentException("Class $templateClass does not extend " . Template::class . ' or it does not exist.');
+		}
+		$this->templateClass = $templateClass;
 	}
 
 }
--- a/tests/Bridges.Latte/TemplateFactory.customTemplate.phpt
+++ b/tests/Bridges.Latte/TemplateFactory.customTemplate.phpt
@@ -1,0 +1,73 @@
+<?php
+
+/**
+ * Test: TemplateFactory custom template
+ */
+
+use Nette\Application\UI;
+use Nette\Bridges\ApplicationLatte\Template;
+use Nette\Bridges\ApplicationLatte\TemplateFactory;
+use Tester\Assert;
+
+
+require __DIR__ . '/../bootstrap.php';
+
+
+class LatteFactoryMock implements Nette\Bridges\ApplicationLatte\ILatteFactory
+{
+	private $engine;
+
+	public function __construct(Latte\Engine $engine)
+	{
+		$this->engine = $engine;
+	}
+
+	public function create()
+	{
+		return $this->engine;
+	}
+}
+
+class TemplateMockWithoutImplement
+{
+
+}
+
+class TemplateMock extends Nette\Bridges\ApplicationLatte\Template
+{
+	private $file = 'ko';
+
+	public function render()
+	{
+		return strrev($this->file);
+	}
+
+	public function setFile($file)
+	{
+		$this->file = $file;
+	}
+
+	public function getFile()
+	{
+		return $this->file;
+	}
+}
+
+
+$factory = new TemplateFactory(new LatteFactoryMock(new Latte\Engine));
+Assert::type(Template::class, $factory->createTemplate());
+
+
+Assert::exception(function () use ($factory) {
+	$factory->setTemplateClass(TemplateMockWithoutImplement::class);
+}, \Nette\InvalidArgumentException::class, 'Class TemplateMockWithoutImplement does not extend Nette\Bridges\ApplicationLatte\Template or it does not exist.');
+
+
+$factory->setTemplateClass(TemplateMock::class);
+$template = $factory->createTemplate();
+Assert::type(TemplateMock::class, $template);
+Assert::type(UI\ITemplate::class, $template);
+Assert::same([], $template->flashes);
+Assert::same('ok', $template->render());
+$template->setFile('bla');
+Assert::same('alb', $template->render());


src/Bridges/ApplicationLatte/TemplateFactory.php
```diff
@@ -31,14 +31,20 @@ class TemplateFactory implements UI\ITemplateFactory
...
 	public function __construct(ILatteFactory $latteFactory, Nette\Http\IRequest $httpRequest = NULL,
-		Nette\Security\User $user = NULL, Nette\Caching\IStorage $cacheStorage = NULL)
+		Nette\Security\User $user = NULL, Nette\Caching\IStorage $cacheStorage = NULL, $templateClassName = Template::class)
 	{
 		$this->latteFactory = $latteFactory;
 		$this->httpRequest = $httpRequest;
 		$this->user = $user;
 		$this->cacheStorage = $cacheStorage;
+
+		\Nette\Utils\Validators::assert($templateClassName, 'string');
```
TomasVotruba: This is not needed, as it is not convention in Nette.
Also PHP 7 is comming soon with type checks.

src/Bridges/ApplicationLatte/TemplateFactory.php
```diff
@@ -48,7 +54,11 @@ public function __construct(ILatteFactory $latteFactory, Nette\Http\IRequest $ht
 	public function createTemplate(UI\Control $control = NULL)
 	{
 		$latte = $this->latteFactory->create();
-		$template = new Template($latte);
+		$template = new $this->templateClassName($latte);
```
TomasVotruba: This looks ok. Just wondering about use case, when custom class has different arguments.

tests/Bridges.Latte/TemplateFactory.customTemplate.phpt
```diff
@@ -0,0 +1,73 @@
...
+
+
+$factory = new TemplateFactory(new LatteFactoryMock(new Latte\Engine), NULL, NULL, NULL, TemplateMock::class);
+$template = $factory->createTemplate();
+Assert::type(TemplateMock::class, $template);
+Assert::type(UI\ITemplate::class, $template);
+Assert::same([], $template->flashes);
+Assert::same('ok', $template->render());
+$template->setFile('bla');
+Assert::same('alb', $template->render());
```
TomasVotruba: Awesome test, thank you.

src/Bridges/ApplicationLatte/TemplateFactory.php
mrtnzlml: It's not convention in Nette doesn't mean I cannot do that, does it? And it's not true that it's not needed (before PHP 7). Moreover PHP 7 is not used in Nette yet.

src/Bridges/ApplicationLatte/TemplateFactory.php
mrtnzlml: It should be in another PR if needed I think. It's not possible to add another arguments with current implementation as well.

tests/Bridges.Latte/TemplateFactory.customTemplate.phpt
```diff
@@ -0,0 +1,73 @@
...
+	}
+}
+
+class TemplateMockWithoutImplement {}
+
+class TemplateMock implements UI\ITemplate
+{
+	private $file = 'ko';
+
+	public function render() {
```
mrtnzlml: TODO: fix coding standards

src/Bridges/ApplicationLatte/TemplateFactory.php
TomasVotruba: Code would be inconsistent with other strings (scalar) types in methods. 

Rather check if this class implements `ITemplate` interface. Because it would fail later and now it silently pass in here creating invalid object.

src/Bridges/ApplicationLatte/TemplateFactory.php
TomasVotruba: Ok, let's skip it now then.

Btw, do you need `$latte` as argument in your implementation?

src/Bridges/ApplicationLatte/TemplateFactory.php
mrtnzlml: I have to do both checks because without checking string it's possible to call `new` on integer for example. From my point of view validation is needed. Can you help me how to prevent calling `$template = new $this->templateClassName($latte);` (line 57) with wrong `$this->templateClassName` type?

TomasVotruba: I see. If so, add them both here. It doesn't make sense splitting them.
Standalone method would be best, e.g. `validateTemplateClassName()`.

src/Bridges/ApplicationLatte/TemplateFactory.php
mrtnzlml: Not necessarily because I am extending `Nette\Bridges\ApplicationLatte\Template` and then calling `$this->getLatte()`. So the answer is no but original implementation needs it.

src/Bridges/ApplicationLatte/TemplateFactory.php
mrtnzlml: Ok, I'll improve it in next iteration. I am used to doing validations in constructor in my projects...

TomasVotruba: Yes. This should be in constructor.

Something like this:

```php
public function __construct($className = Tempalte::class)
{
    $this->validateTempalteClassName($className);
}
```

mrtnzlml: I don't see benefit here. It's only because of calling line `Nette\Utils\Validators::assert($templateClassName, 'string');` here. I don't like it... :)

Waiting for more opinions...

TomasVotruba: We missunderstood a bit. I've updated the code.

mrtnzlml: Yeah, it cannot be done like that. It's because `$template` is created during `createTemplate` call (everytime you call it) and therefore you cannot test its instance in constructor. You don't have `ITemplate` in constructor. What implements `ITemplate`? The answer is: created `$template`.

JanTvrdik:
Setter (as suggested in #141) is IMHO better (template class name has a character of *configuration* not of *dependency*)

mrtnzlml:
@JanTvrdik I love [this idea](https://github.com/nette/application/issues/141#issuecomment-249413992) (despite the naming - just idea) even more.

JanTvrdik:
You mean the „TemplateClassFactory“ which would create just instance of `Template` without configuring it any further? That seems like a terrible idea to me.

Good solutions are either setter or splitting the instance creation and configuration to multiple methods to allow overriding part of the logic.

TomasVotruba:
**I agree with @JanTvrdik in setter as configuration.**

If you want to change class name, you might want to do it multiple times. With constructor dependency you would have to create new instance of TemplateFactorry.

mrtnzlml:
TODO for me: Change implementation back to the setter one. Make it configurable using DI extension if possible (because original usage is not nice).

mrtnzlml:
Ok guys, what about this implementation?

src/Bridges/ApplicationDI/LatteExtension.php
```diff
@@ -19,6 +19,7 @@ class LatteExtension extends Nette\DI\CompilerExtension
 	public $defaults = [
 		'xhtml' => FALSE,
 		'macros' => [],
+		'templateClass' => \Nette\Bridges\ApplicationLatte\Template::class
```
mrtnzlml: Is this coupling between bridges fine?

src/Bridges/ApplicationLatte/TemplateFactory.php
```diff
@@ -48,7 +51,11 @@ public function __construct(ILatteFactory $latteFactory, Nette\Http\IRequest $ht
 	public function createTemplate(UI\Control $control = NULL)
 	{
 		$latte = $this->latteFactory->create();
-		$template = new Template($latte);
+		$template = new $this->templateClass($latte);
+		if (!$template instanceof UI\ITemplate) {
```
JanTvrdik: This check should be moved to `setTemplateClass` (see is_a with third argument set to TRUE)

src/Bridges/ApplicationLatte/TemplateFactory.php
```diff
@@ -112,4 +119,10 @@ public function createTemplate(UI\Control $control = NULL)
 		return $template;
 	}
 
+	public function changeTemplateClass($templateClass)
```
JanTvrdik: should be named setTemplateClass (due to consistency)

src/Bridges/ApplicationLatte/TemplateFactory.php
```diff
@@ -112,4 +119,10 @@ public function createTemplate(UI\Control $control = NULL)
...
+	{
+		\Nette\Utils\Validators::assert($templateClass, 'string');
```
JanTvrdik: should be removed (due to consistency)

src/Bridges/ApplicationDI/LatteExtension.php
JanTvrdik: * should be simplified to `Nette\Bridges\ApplicationLatte\Template::class`
* the usecase se imho too narrow to add option to the extension so I would remove this entirely
* if added, the default value should be `NULL` and the setter call should only be emitted if the value is not NULL

src/Bridges/ApplicationLatte/TemplateFactory.php
JanTvrdik: You will also probably need to call `class_exists` on the `$templateClass` to load the class to memory

src/Bridges/ApplicationLatte/TemplateFactory.php
mrtnzlml: How would you prevent passing something different than string? I understand your note about consistency, but it's wrong I think. Or should I remove usage of `Validators` and use custom check?

src/Bridges/ApplicationLatte/TemplateFactory.php
mrtnzlml: Thanks - I didn't know the third argument of is_a.

mrtnzlml:
Updated - I have no idea why is old conversation still here (what a mess in this discussion)... :-/
What changed:

- it uses setter implementation
- it's still configurable via Latte DI extension
- it's still used validation in `setTemplateClass` because I strongly believe that we should check if passed class name is string (and not created instance for example)

Thanks for help.

src/Bridges/ApplicationLatte/TemplateFactory.php
JanTvrdik: > How would you prevent passing something different than string? 

Here is the secret – you don't prevent passing something different because this is PHP because it would kill code readability and performance. Pretty much every Nette method without typehint can be broken by passing the wrong type.

src/Bridges/ApplicationLatte/TemplateFactory.php
```diff
@@ -112,4 +115,13 @@ public function createTemplate(UI\Control $control = NULL)
 		return $template;
 	}
 
+	public function setTemplateClass($templateClass)
+	{
+		\Nette\Utils\Validators::assert($templateClass, 'string');
+		if (!is_a($templateClass, UI\ITemplate::class, TRUE)) {
+			throw new \Nette\Utils\AssertionException('Template should be instance of ' . UI\ITemplate::class . ", $templateClass given.");
```
JanTvrdik: * should be Nette\InvalidArgumentException, imho
* you need to call `!class_exist($templateClass) || ` before `!is_a` because `is_a` does not trigger autoloading

src/Bridges/ApplicationDI/LatteExtension.php
```diff
@@ -52,9 +53,12 @@ public function loadConfiguration()
 			->addSetup('Nette\Utils\Html::$xhtml = ?', [(bool) $config['xhtml']])
 			->setImplement(Nette\Bridges\ApplicationLatte\ILatteFactory::class);
 
-		$builder->addDefinition($this->prefix('templateFactory'))
+		$templateFactory = $builder->addDefinition($this->prefix('templateFactory'))
 			->setClass(Nette\Application\UI\ITemplateFactory::class)
 			->setFactory(Nette\Bridges\ApplicationLatte\TemplateFactory::class);
+		if($config['templateClass'] !== NULL) {
```
JanTvrdik: missing space after if

mrtnzlml:
Updated according to comments.

src/Bridges/ApplicationLatte/TemplateFactory.php
```diff
@@ -112,4 +115,12 @@ public function createTemplate(UI\Control $control = NULL)
 		return $template;
 	}
 
+	public function setTemplateClass($templateClass)
+	{
+		if (!class_exists($templateClass) || !is_a($templateClass, UI\ITemplate::class, TRUE)) {
+			throw new \Nette\InvalidArgumentException('Template should be instance of ' . UI\ITemplate::class . ", $templateClass given.");
```
JanTvrdik: * `Nette\InvalidArgumentException`
* "Class '$templateClass' does not implement ... ."

src/Bridges/ApplicationLatte/TemplateFactory.php
```diff
@@ -112,4 +115,12 @@ public function createTemplate(UI\Control $control = NULL)
 		return $template;
 	}
 
+	public function setTemplateClass($templateClass)
+	{
+		if (!class_exists($templateClass) || !is_a($templateClass, UI\ITemplate::class, TRUE)) {
```
dg: It should be instance of `Template::class`

mrtnzlml: Why not `ITemplate`? Can you please elaborate it more @dg? I don't fully understand... :) Thank you.

JanTvrdik: My guess is that it's because `TemplateFactory` has `@return Template` annotation. However Nette itself does not seem to rely on the instance being `Template`.

dg: It has `@return Template`, it assumes `__construct(Latte\Engine)` and `__set` for variables (or fields).

mrtnzlml:
Rebased on top of master branch (because of failing tests).

mrtnzlml:
Thank you @dg! :)

PR #161, opened 2016-09-29
Route: fixed bug in Route::constructUrl()

vlastavesely:
Fix for: https://forum.nette.org/en/26382-a-bug-in-route-constructurl


dg:
Thanks

PR #162, opened 2016-10-03
RoutingExtension: added support for custom single-route classes.

vlastavesely:
An alternative way to add new Route to RouteList. It moves control over the Route class into the RouteList, so Neon configuration for RoutingExtension can be used by other combination of classes descended from Nette*\RouteList and Nette*\Route.

### DIFF
--- a/src/Bridges/ApplicationDI/RoutingExtension.php
+++ b/src/Bridges/ApplicationDI/RoutingExtension.php
@@ -19,6 +19,7 @@ class RoutingExtension extends Nette\DI\CompilerExtension
 	public $defaults = [
 		'debugger' => NULL,
 		'routes' => [], // of [mask => action]
+		'routeClass' => NULL,
 		'cache' => FALSE,
 	];
 
@@ -26,7 +27,7 @@ class RoutingExtension extends Nette\DI\CompilerExtension
 	private $debugMode;
 
 
-	public function __construct($debugMode = FALSE)
+	public function __construct(bool $debugMode = FALSE)
 	{
 		$this->defaults['debugger'] = interface_exists(Tracy\IBarPanel::class);
 		$this->debugMode = $debugMode;
@@ -42,8 +43,10 @@ class RoutingExtension extends Nette\DI\CompilerExtension
 			->setClass(Nette\Application\IRouter::class)
 			->setFactory(Nette\Application\Routers\RouteList::class);
 
-		foreach ($config['routes'] as $mask => $action) {
-			$router->addSetup('$service[] = new Nette\Application\Routers\Route(?, ?);', [$mask, $action]);
+		$routeClass = $this->config['routeClass'] ?: 'Nette\Application\Routers\Route';
+
+		foreach ($this->config['routes'] as $mask => $action) {
+			$router->addSetup('$service[] = new ' . $routeClass . '(?, ?)', [$mask, $action]);
 		}
 
 		if ($this->name === 'routing') {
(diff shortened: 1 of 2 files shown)


dg:
What is it good for?

vlastavesely:
With this, complex "drop-in" routers can be made using existing declaration for RoutingExtension.

Now, when the 'routing.router' is replaced by other class that depends e.g. on a translator, its children remain to be Nette\\*\Route that does not know what to do with that translator. You cannot set filters to $metadata in config.neon, so you have to replace routes with class that does it inside itself.

If my PR accepted, there is possible to simply override addRoute() where child can get its dependencies - not even for translation but e. g. for DB filtering, etc. This solution keeps freedom to declare routes both in bootstrap and config.neon. Otherwise, configuration for RoutingExtension could not be used in this case and everything would have to be done in the bootstrap.php.

vlastavesely:
*I meant "drop-in" replacement of course.

rostenkowski:
:+1:

dg:
So it is not about RouteList, it is about RoutingExtension. I see RoutingExtension as experimental and I welcome any improvements. These improvements, that solves your use case, should be IMHO done in RoutingExtension.

vlastavesely:
I do agree. After your suggestion I've tried to find some better solution and I think I've found one.

In many cases (even in Nette and my project) RouteList and Route will be located in the same namespace. So, it would be sufficient to create routes from class Route located in the same namespace the RouteList belongs to. It not only solves my problem but also it allows me to leave my config.neon without any changes. To make it even more useful, I have added possibility to change child route's class totally by parameter routing.routeClass.

This solution has, though, one limitation. It cannot pass dependencies via constructor to child routes. These have to be set later by calling e.g. Route::setTranslator(). But this is not problem for functionality.

vlastavesely:
Updated. Class Route from RouteList's namespace is used only if exists.

src/Bridges/ApplicationDI/RoutingExtension.php
```diff
@@ -55,6 +52,21 @@ public function loadConfiguration()
 	public function beforeCompile()
 	{
 		$builder = $this->getContainerBuilder();
+		$router = $builder->getDefinition($this->prefix('router'));
+
+		if ($this->config['routeClass']) {
+			$class = $this->config['routeClass'];
+		} else {
+			$namespace = Nette\PhpGenerator\Helpers::extractNamespace($router->getFactory()->getEntity());
```
vlastavesely: In the case that a class `Route` exists in the same NS as RouteListm, yes. But this magic could be removed and used `$this->config['routeClass']` if it is set or `Nette\Application\Routers\Route::class` otherwise. In that case, there should not be any BC.

dg: So can you change it?

vlastavesely:
done

src/Bridges/ApplicationDI/RoutingExtension.php
```diff
@@ -55,6 +52,13 @@ public function loadConfiguration()
 	public function beforeCompile()
 	{
 		$builder = $this->getContainerBuilder();
+		$router = $builder->getDefinition($this->prefix('router'));
```
dg: $router is not needed, or is it? It can be moved back to loadConfiguration() IMHO.

vlastavesely: :( yes, you're right. it had reason in previous version but after the last changes, it can be moved back into loadConfiguration. I am gonna change it withing a few minutes.

dg: Thanks

vlastavesely:
done

dg:
Thanks

PR #163, opened 2016-10-14
Application: fixed empty latest request in request to ErrorPresenter when createInitialRequest() failed.

vlastavesely:
When `Application::createInitialRequest()` ends with an exception (missing presenter), `Application::processException()` is called before `Application::processRequest()`. This is problem for `processException()` tries to read property `Application::$requests` that is, however, set later by `processRequest()`. So, even if application got request from router, `ErrorPresenter` gots `NULL` instead of that request.

For this one specific case it should be set manually.

### DIFF
--- a/src/Application/Application.php
+++ b/src/Application/Application.php
@@ -124,7 +124,7 @@ class Application
 			$name = $request->getPresenterName();
 			$this->presenterFactory->getPresenterClass($name);
 		} catch (InvalidPresenterException $e) {
-			throw new BadRequestException($e->getMessage(), 0, $e);
+			throw new BadRequestException($e->getMessage(), 0, $e, $request);
 		}
 
 		return $request;
@@ -165,14 +165,22 @@ class Application
 	 */
 	public function processException($e)
 	{
-		if (!$e instanceof BadRequestException && $this->httpResponse instanceof Nette\Http\Response) {
-			$this->httpResponse->warnOnBuffer = FALSE;
-		}
-		if (!$this->httpResponse->isSent()) {
-			$this->httpResponse->setCode($e instanceof BadRequestException ? ($e->getCode() ?: 404) : 500);
+		if ($e instanceof BadRequestException) {
+			$httpCode = $e->getCode() ?: 404;
+			$lastRequest = $e->getRequest() ?: end($this->requests);
+		} else {
+			$httpCode = 500;
+			$lastRequest = end($this->requests);
+			if ($this->httpResponse instanceof Nette\Http\Response) {
+				$this->httpResponse->warnOnBuffer = FALSE;
+			}
 		}
 
-		$args = ['exception' => $e, 'request' => end($this->requests) ?: NULL];
+		if (!$this->httpResponse->isSent()) {
+			$this->httpResponse->setCode($httpCode);
+		}
+
+		$args = ['exception' => $e, 'request' => $lastRequest];
 		if ($this->presenter instanceof UI\Presenter) {
 			try {
 				$this->presenter->forward(":$this->errorPresenter:", $args);
--- a/src/Application/exceptions.php
+++ b/src/Application/exceptions.php
@@ -41,12 +41,24 @@ class BadRequestException extends \Exception
 	/** @var int */
 	protected $code = 404;
 
+	/** @var Request|NULL */
+	protected $request;
 
-	public function __construct($message = '', $code = 0, \Exception $previous = NULL)
+
+	public function __construct($message = '', $code = 0, \Exception $previous = NULL, Request $request = NULL)
 	{
 		parent::__construct($message, $code < 200 || $code > 504 ? $this->code : $code, $previous);
+		$this->request = $request;
 	}
 
+
+	/**
+	 * @return Request|NULL
+	 */
+	public function getRequest()
+	{
+		return $this->request;
+	}
 }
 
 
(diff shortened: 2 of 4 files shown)


TomasVotruba:
Could you please add test for that as well?

vlastavesely:
Test added.

dg:
This should be fixed in different way, because createInitialRequest should not sometimes change $requests as side effect.

TomasVotruba:
Any suggestions?

vlastavesely:
I'm thinking about that. I've found two another solutions but they are not as direct as the first one.

One is to call Router::match() before Application::createInitialRequest(), save it into variable and pass it as second parameter into Application::processException() where it can be used if end($this->requests) is NULL.

```
	$request = $this->router->match($this->httpRequest);
	$this->processRequest($this->createInitialRequest($request));
	
	...
	
	$this->processException($e, $request); // in Application::run(), if exception thrown
	
	...
	
	$args = ['exception' => $e, 'request' => end($this->requests) ?: $request]; // in processException()
```


The second solution is simply to call Router::match() again if last response is NULL like that:
	
```
	$request = end($this->requests);
	if (!$request) {
		$request = $this->router->match($this->httpRequest);
	}
	$args = ['exception' => $e, 'request' => $request];
```

In the second solution there is unnecessary calling Route::match() again if createInitialReqiest() fails (we already had apprequest from router so why call it again?), first solution is more similar to my original suggestion but not as direct as that one.

JanTvrdik:
I think that you should store the request object inside `BadRequestException` and `processException`  should get the request object from `BadRequestException` if available.

vlastavesely:
Updated. Thanks for suggestion!

src/Application/exceptions.php
```diff
@@ -41,12 +41,23 @@ class BadRequestException extends \Exception
...
-	public function __construct($message = '', $code = 0, \Exception $previous = NULL)
+
+	public function __construct($message = '', $code = 0, \Exception $previous = NULL, Request $request = NULL)
 	{
 		parent::__construct($message, $code < 200 || $code > 504 ? $this->code : $code, $previous);
+		$this->request = $request;
 	}
 
+	/**
+	 * @return Request
```
JanTvrdik: `@return Request|NULL`

src/Application/exceptions.php
```diff
@@ -41,12 +41,23 @@ class BadRequestException extends \Exception
 	/** @var int */
 	protected $code = 404;
 
+	/** @var Request */
```
JanTvrdik: `Request|NULL`

src/Application/exceptions.php
```diff
@@ -41,12 +41,23 @@ class BadRequestException extends \Exception
...
 
-	public function __construct($message = '', $code = 0, \Exception $previous = NULL)
+
+	public function __construct($message = '', $code = 0, \Exception $previous = NULL, Request $request = NULL)
 	{
 		parent::__construct($message, $code < 200 || $code > 504 ? $this->code : $code, $previous);
+		$this->request = $request;
 	}
 
+	/**
```
JanTvrdik: there should always be two empty lines between methods

src/Application/Application.php
```diff
@@ -172,7 +172,7 @@ public function processException($e)
 			$this->httpResponse->setCode($e instanceof BadRequestException ? ($e->getCode() ?: 404) : 500);
 		}
 
-		$args = ['exception' => $e, 'request' => end($this->requests) ?: NULL];
+		$args = ['exception' => $e, 'request' => end($this->requests) ?: $e->getRequest()];
```
JanTvrdik: This assumes that `$this->requests` can only be empty when `BadRequestException` was thrown which I'm not sure is something we can rely on.

What about rewriting the whole method like this?

~~~php
if ($e instanceof BadRequestException) {
	$httpCode = $e->getCode() ?: 404;
	$lastRequest = $e->getRequest() ?: end($this->requests);

} else {
	$httpCode = 500;
	$lastRequest = end($this->requests);

	if ($this->httpResponse instanceof Nette\Http\Response) {
		$this->httpResponse->warnOnBuffer = FALSE;
	}
}

$this->httpResponse->setCode($httpCode);
$params = ['exception' => $e, 'request' => $lastRequest];

...
~~~

vlastavesely: Hm, yes. If Router::match() fails, it can make problems. That check of type of $e is needed.

dg:
What about this? https://github.com/dg/nette-application/commit/022574b5b7bac770b093d84a0c6f640d4f5f8821

vlastavesely:
It solves one problem but creates another. If no route matches, it ignores Application::$errorPresenter and shows Nette's default. Right now, I do not know where is problem, I'll check it later.

With today's solution in my PR it works correctly (in all tested cases there is custom ErrorPresenter).

dg:
fixed https://github.com/dg/nette-application/commit/0ca909743f51f31236231959166043e6ea10cc39

vlastavesely:
`throw $isForward ? new BadRequestException($e->getMessage(), 0, $e) : $e;`

This line makes problem. It throws InvalidPresenterException, so error code is 500. I think that for missing presenter would be better 404. What about throw always BadRequestException?

dg:
Yes, I noticed it. https://github.com/dg/nette-application/commit/967890567ed6ef09907e7ce12ee1d4993dc29149

vlastavesely:
OK, now it seems to be working in all cases.

TomasVotruba:
Great job, thank you guys

PR #167, opened 2016-12-10
Route::constructUrl(): split of module and presenter moved before global filter

vlastavesely:
I was doing some changes in structure of my application and I have found something what is IMHO bug.

In short: right now, the global input filter gets module and presenter split (like this: [Module: 'Front', Presenter: 'Homepage']) but output filter gets them together as presenter ([Presenter: 'Front:Homepage']).

Wouldn't it be better to unify the behaviour of those two functions to get the params split in both cases?


dg:
Makes sense, thanks.

PR #169, opened 2017-01-17
RoutingPanel: fixed whitespaces in routelist

vlastavesely:
After commit `ec3380b7` in Tracy, routing panel looks ugly. Top image is how it looks now, the second is how it looks like after my change.

bug


dg:
Thanks. It is bug in Tracy, I fixed it there.

PR #176, opened 2017-02-23
Removed TABs from routing panel

meridius:
- bug fix? **yes**
- new feature? **no**
- BC break? **no**

Because of indentation within the `<code>` tags the Tracy routing panel is rendering TAB characters in data of *Defaults* and *Matched as* columns. Thus making its content full of white space, lengthy and hard to read due to needed scrolling. 

This PR fixes the described issue by rewriting rendering of the problematic `<code>` tags.  
The second commit just removes unnecessary (in my opinion) PHP tags, making given section easier to read.

An finally, even if this PR adds a few code lines, I believe it is for better readability and the rest of the file should be rewritten in the same fashion someday.


dg:
How I can reproduce the problem?

meridius:
You don't see it? That's weird. It is happening on a clean `composer create-project nette/web-project` for me.

dg:
Can you send screenshot?

meridius:
http://imgur.com/a/UuYtt

meridius:
And with text selected http://imgur.com/a/wRcCX

dg:
Try update Tracy

meridius:
Heh, fun. 
Fixed, thanks 😄

PR #179, opened 2017-04-12
Presenter: set defaults to $globalParams, fixes foreach over NULL

TomasVotruba:
Breaks on `getGlobalState()`, where is [foreached null](https://github.com/nette/application/blob/d0f19faf7301197049ccd9fa9eda9419e85d9ea5/src/Application/UI/Presenter.php#L1074).

```php
protected function getGlobalState($forClass = NULL): array
{
	$sinces = &$this->globalStateSinces;
	if ($this->globalState === NULL) {
		$state = [];
		// foreach over NULL
		foreach ($this->globalParams as $id => $params) {
			$prefix = $id . self::NAME_SEPARATOR;
			foreach ($params as $key => $val) {
			// ...
```

Ii think t's called in layout resolution in [`$template`](https://github.com/nette/application/blob/d0f19faf7301197049ccd9fa9eda9419e85d9ea5/src/Bridges/ApplicationLatte/TemplateFactory.php#L113) by provider.

```php
if ($control) {
	$template->control = $control;
	$template->presenter = $presenter;

        // here
	$latte->addProvider('uiControl', $control);
	$latte->addProvider('uiPresenter', $presenter);

	// ...
}
```

- bug fix? yes/no   
- new feature? yes/no
- BC break? yes/no
- doc PR: nette/docs#???

### DIFF
--- a/src/Application/UI/Presenter.php
+++ b/src/Application/UI/Presenter.php
@@ -63,7 +63,7 @@ abstract class Presenter extends Control implements Application\IPresenter
 	public $absoluteUrls = FALSE;
 
 	/** @var array */
-	private $globalParams;
+	private $globalParams = [];
 
 	/** @var array */
 	private $globalState;


dg:
Can you find you why is `getGlobalState` called before it is initialized? What is debug stack trace?

michalhlavka:
@dg I've got same issue. I am generating link in function which is called by config.

TomasVotruba:
N
TnksNot relevant anymore, nor time or intention to reproduce.

PR #182, opened 2017-06-07
[WIP] Add nette/coding-standard

TomasVotruba:
Based on discussion with @dg , this the best repo to start with.

## Goals

- identify useful checkers
- drop the rest
- see, how it works as dependency package

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,9 +32,7 @@
 		"nette/security": "^3.0",
 		"latte/latte": "^3.0",
 		"tracy/tracy": "^2.4",
-		"mockery/mockery": "^1.0",
-		"nette/coding-standard": "@dev",
-		"symplify/easy-coding-standard": "@dev"
+		"mockery/mockery": "^1.0"
 	},
 	"conflict": {
 		"nette/nette": "<2.2",
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,8 @@
 		"latte/latte": "^3.0",
 		"tracy/tracy": "^2.4",
 		"mockery/mockery": "^1.0",
-		"nette/coding-standard": "dev-master"
+		"nette/coding-standard": "@dev",
+		"symplify/easy-coding-standard": "@dev"
 	},
 	"conflict": {
 		"nette/nette": "<2.2",
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,8 @@
 		"nette/security": "^3.0",
 		"latte/latte": "^3.0",
 		"tracy/tracy": "^2.4",
-		"mockery/mockery": "^1.0"
+		"mockery/mockery": "^1.0",
+		"nette/coding-standard": "dev-master"
 	},
 	"conflict": {
 		"nette/nette": "<2.2",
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Does anyone have idea why dependencies are not installed? 

Probably many dev dependencies.

This waits on https://github.com/squizlabs/PHP_CodeSniffer/pull/1442#issuecomment-302253910 then

dg:
It must be installed the same way as nette codechecker, adding as composer dependency is bad way.

TomasVotruba:
@dg Could you elaborate, why do you think so?

If somebody will make a PR, how could he fix the code locally then?

I'll look on this after CodeSniffer 3.0.1 is released.

composer.json
```diff
@@ -32,7 +32,9 @@
 		"nette/security": "^3.0",
 		"latte/latte": "^3.0",
 		"tracy/tracy": "^2.4",
-		"mockery/mockery": "^1.0"
+		"mockery/mockery": "^1.0",
+		"nette/coding-standard": "@dev",
+		"symplify/easy-coding-standard": "@dev"
```
f3l1x: `symplify/easy-coding-standard` is also needed? I thought it's in nette/coding-standard repo, isn't it?

TomasVotruba: Have you tried installing "nette/coding-standard" as dependency? I think it should fail without this.

dg:
The developer just installs nette/coding-standard locally and runs it.

TomasVotruba:
@dg What is the advantage of this?

dg:
This is way how to install and run easy-coding-standard (but it currently fails on autoloading error).
```
composer create-project symplify/easy-coding-standard --no-dev
easy-coding-standard\bin\easy-coding-standard check src
```

dg:
It is not about advantage, it is the only possible solution how to install it. Because easy-coding-standard is separate tool, it is not part of nette/application. It also uses nette/application, so there is no way how to install nette/application into nette/application.

TomasVotruba:
I see, thanks for explaining.

I have one last blind spot. If I install it locally to the package like this, how do I make sure I won't accidentally commit it?

dg:
> If I install it locally to the package like this, how do I make sure I won't accidentally commit it?

Of course, it's better to install it in another directory. I can develop more nette libraries and it makes no sense to install easy-coding-standard multiple times. But if I want to have it in library folder, I'd add a line to `.git/info/exclude`.

>  I wonder what is application actually needed for. Probably not.

It's not just about `application`, you should not use any library from nette.

PhpStan is similarly broken, so it cannot be used for Nette or Symfony.

TomasVotruba:
I see. I wish there was more elegant solution than installing a project and ignoring it.

dg:
Possible workaround is to install it into `vendor`, because it is ignored by all libraries `composer create-project symplify/easy-coding-standard vendor/easy-coding-standard --no-dev`

But I wouldn't want have it in the project at all, because I it should be hidden for PhpStorm too etc.

TomasVotruba:
This can be closed, right?

PR #186, opened 2017-08-28
docs: fix invalid annotation

vlastavesely:


src/Application/UI/ComponentReflection.php
```diff
@@ -231,7 +231,7 @@ public static function parseAnnotation(\Reflector $ref, $name): ?array
 
 
 	/**
-	 * @return [string, bool]
+	 * @return mixed[]
```
JanTvrdik: It's much better to use `array` than `mixed[]` (because `<type>[]` is unfortunately not used consistently by PHP community and resolving it's meaning accurately requires quite complex heuristics). 

Also you've dropped the information about the structure of the array. It would be better to do sth. like

~~~php
/**
 * @return array [string, bool]
 */
~~~

PR #187, opened 2017-08-29
some classes & members marked as final (BC break)

dg:
This among others reverts commit 45ed648caeb7dd94c6f38bd50d7526606a7bcc14

- BC break? yes

### DIFF
--- a/src/Application/Application.php
+++ b/src/Application/Application.php
@@ -170,7 +170,7 @@ class Application
 	 * Returns all processed requests.
 	 * @return Request[]
 	 */
-	public function getRequests(): array
+	final public function getRequests(): array
 	{
 		return $this->requests;
 	}
@@ -179,7 +179,7 @@ class Application
 	/**
 	 * Returns current presenter.
 	 */
-	public function getPresenter(): ?IPresenter
+	final public function getPresenter(): ?IPresenter
 	{
 		return $this->presenter;
 	}
--- a/src/Application/ErrorPresenter.php
+++ b/src/Application/ErrorPresenter.php
@@ -17,7 +17,7 @@ use Tracy\ILogger;
 /**
  * Default Error Presenter.
  */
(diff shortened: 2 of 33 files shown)


TomasVotruba:
This might lead to some magic remove. Great job :+1: 

Could :-1:s add some comments with reasoning please?

TomasVotruba:
Well, we can:

* https://phpfashion.com/how-to-mock-final-classes
* https://www.tomasvotruba.cz/blog/2019/03/28/how-to-mock-final-classes-in-phpunit/

lukasojd:
-1 again

https://github.com/nette/security/commit/c7cc80bbe5f8db6346b7008f88eb3b478380d6cd

lukasojd:
> @lukasojd Well, we can:
> 
>     * https://phpfashion.com/how-to-mock-final-classes
> 
>     * https://www.tomasvotruba.cz/blog/2019/03/28/how-to-mock-final-classes-in-phpunit/

Thank you. Final is evel and I will not use dg/bypass-finals. Author commits is author package. DG need more money and he insert keyword final everywhere.

mabar:
> DG need more money and he insert keyword final everywhere.

That's very dumb claim, you know? Please stop being an asshole complaining everywhere about few final getters because you somehow didn't get how to solve problem even after I explained it to you several times. How could you complain about bypassing finals and be okay with mocks same time? Both is bad magic to solve edge cases and this is really not the case in which you really need to use it.

janfejtek:
My use case is that I have custom Authenticator, User and Identity classes. 
Originally I override getUser in presenter and it was returning CustomUser class (basically just hinting PhpStorm). But now, PhpStorm is showing errors about missing methods. Because I had to remove getUser from my presenter as well as getIdentity from CustomUser
PHP 7.4 supports covariance so more specific return type should not be a problem in overridden method

Or am I doing it wrong?

PR #189, opened 2017-09-13
Remove typehint for $message in flashMessage

venca-x:
I'm removing typehint for display Nette\Utils\Html in $message

- bug fix? no
- new feature? no
- BC break? no

### DIFF
--- a/src/Application/UI/Control.php
+++ b/src/Application/UI/Control.php
@@ -69,7 +69,7 @@ abstract class Control extends Component implements IRenderable
 	/**
 	 * Saves the message to template, that can be displayed after redirect.
 	 */
-	public function flashMessage(string $message, string $type = 'info'): \stdClass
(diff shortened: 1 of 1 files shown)


CZechBoY:
You should add additional check for type Html|string and throw exception if it is not.

dg:
Thx

JanTvrdik:
It would be nice to add a test specifically for `Nette\Utils\Html` to make sure it continues to work in the future.

Adding a typehint for `Nette\Utils\Html|string` is also a possibility but depends on whether the intention is to allow only `Nette\Utils\Html` instances or any instances with `__toString` method is fine. What do you think @dg?

dg:
I would leave it completely free. It is like template parameter.

CZechBoY:
@milo so you are using more parameters than this definition and you have your own template implementation, right?

milo:
I was planning to write template macro but I stuck with linked trait due to lack of time.

PR #190, opened 2018-01-13
Removed goto from Application::processRequest

lysektomas:
Hi, I have removed goto from Application::processRequest method. I know, that some people may say that this is a cosmetic change, but, in my opinion, using goto's in the 21st century is heresy. (http://php.net/manual/en/control-structures.goto.php )

Before I changed the code, I have added few test cases for better mutation test result of this function.

- bug fix? no   
- new feature? no
- BC break? no

### DIFF
--- a/src/Application/Application.php
+++ b/src/Application/Application.php
@@ -124,7 +124,27 @@ class Application
 	 */
 	public function processRequest(Request $request)
 	{
-		$response = $this->getResponseFromRequest($request);
+		do{
+			if (count($this->requests) > self::$maxLoop) {
+				throw new ApplicationException('Too many loops detected in application life cycle.');
+			}
+
+			$this->requests[] = $request;
+			$this->onRequest($this, $request);
+
+			if (!$request->isMethod($request::FORWARD) && !strcasecmp($request->getPresenterName(), $this->errorPresenter)) {
+				throw new BadRequestException('Invalid request. Presenter is not achievable.');
+			}
+
+			try {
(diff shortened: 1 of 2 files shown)


matej21:
you can see related discussion here: https://github.com/nette/application/commit/80598ab1bb680e392c41eec8275051db8d9c487f

lysektomas:
OK, I see. I will try to reimplement recursion to iteration. Will it be an acceptable solution? If not, those tests could be used anyway.

dg:
If the only reason is `using goto's in the 21st century is heresy`, please close it. It is FUD.

lysektomas:
What is FUD?

JanTvrdik:
https://en.wikipedia.org/wiki/Fear,_uncertainty_and_doubt

lysektomas:
A few years back I have read on Nette website that Nette comes from French word clean/clear or from your site:  "Nette puts extra emphasis productivity, **clean code**, and security. " 

And you are still willingly using goto statements? Are you considering goto as a clean way of making _something_? 

Where is a problem? Do you think that there is some level of uncertainty or doubt when you are not using one of the worst construction in every programing language? Do you disagree with Djikstra's remarks or do you disagree with a way how programming is evolving in last 30 years? 

How is it even possible to use goto statement in the object-oriented design? Is it a joke, easter-egg or something internal?

dg:
I will be glad if you send another PR, but I do not want to merge it because I really think `goto` is a clean way. You can consider it a easter-egg.

PR #191, opened 2018-02-05
FileResponse: strict type fix

vlastavesely:
Since `4e6` is float and function `fread()` exects integer value for
length, it caused a `TypeError` on bigger files.

- bug fix? yes
- BC break? no


dg:
Thanks!

PR #194, opened 2018-03-11
Override $presenter annotation for better IDE hint

mabar:
- bug fix? no
- new feature? yes
- BC break? no


mabar:
It is probably not useful at all (if you don't use $this->presenter in presenter for any weird reasons). I found it randomly thanks to phpstan and refactoring.

dg:
So why are you sending useless PR?

mabar:
Probably just a bad day, sorry. I found something what was not working as expected and didn't think about it is extremly rare and illogical use case. Feel free to close that.

PR #202, opened 2018-10-04
Fix warning while forwarding to Error presenter

xificurk:
- bug fix
- BC break? no
- doc PR: not needed

`Application::processException()` in some cases uses presenter instance to create the forwarding request to Error presenter, see https://github.com/nette/application/blob/master/src/Application/Application.php#L159
The problem is that at this point we're not guaranteed that the presenter has intialized global state (namely `Presenter::$globalParams` might be null), e.g. the `run()` method was not yet called. This results in pretty confusing warning:
> Warning: Invalid argument supplied for foreach() in /var/www/vendor/nette/application/src/Application/UI/Presenter.php:1143

The easiest fix is to set the default value for `globalParams` to empty array.


dg:
Thanks!

PR #205, opened 2018-11-29
Add getter for httpRequest to application

pavelmlejnek:
- new feature
- BC break: no
- doc PR: not needed

Sometimes I need to work with Nette\Http\IRequest inside an event (mainly onStartup), but there is a getter missing, so I have to do it like that:

```php
$reflectionClass = new \ReflectionClass(Application::class);
$httpRequestProperty = $reflectionClass->getProperty('httpRequest');
$httpRequestProperty->setAccessible(true);

$httpRequest = $httpRequestProperty->getValue();
```

### DIFF
--- a/src/Application/Application.php
+++ b/src/Application/Application.php
@@ -198,6 +198,15 @@ class Application
 
 
 	/**
+	 * Returns HTTP request.
+	 */
+	public function getHttpRequest(): Nette\Http\IRequest
+	{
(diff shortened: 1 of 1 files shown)


mabar:
Why can't you get IRequest from DIC, as other services?

pavelmlejnek:
Yep, that's how I solved it in the end.

PR #206, opened 2018-12-10
Added $templateFile to UI\Control

holantomas:
- new feature
- BC break? no (3rd party apps maybe ...?)
- doc PR: not needed

I made topic for this on forum: https://forum.nette.org/en/31591-ui-control-templatefile

### DIFF
--- a/src/Application/UI/Control.php
+++ b/src/Application/UI/Control.php
@@ -26,7 +26,7 @@ abstract class Control extends Component implements IRenderable
 	private $templateFactory;
 
 	/** @var string|null */
-	private $templateFile = null;
+	private $templateFile;
 
 	/** @var ITemplate */
 	private $template;
@@ -52,7 +52,6 @@ abstract class Control extends Component implements IRenderable
 		if ($this->template !== null) {
 			$this->template->setFile($templateFile);
 		}
-
 		return $this;
 	}
 
--- a/src/Application/UI/Control.php
+++ b/src/Application/UI/Control.php
@@ -25,6 +25,9 @@ abstract class Control extends Component implements IRenderable
 	/** @var ITemplateFactory */
 	private $templateFactory;
 
+	/** @var string|null */
+	private $templateFile = null;
+
 	/** @var ITemplate */
 	private $template;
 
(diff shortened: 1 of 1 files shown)


mabar:
I prefer a higher level abstraction which would use views instead of specific templates, similarly like presenters do.

src/Application/UI/Control.php
```diff
@@ -42,6 +45,23 @@
 	}
 
 
+	final public function setTemplateFile(string $templateFile = null)
+	{
+		$this->templateFile = $templateFile;
+
+		if ($this->template !== null) {
```
mabar: Conditional setter? Never seen that before. Should throw exception or create template if not already created.
I understand why it's here, but it's a just a very specific usecase.

holantomas: I don't know if you got it right. Whole idea of this PR is to be able setup file without creating template. If we throw exception or create template there is no PR needed any more, cause it will work just like shortcut for `$control->getTemplate()->setFile()`. 

But I thought about it if I should add this condition. Because it's just for specific usecase as you said when user try to set file after template was created to prevent "WTF?" situations and let setter work like user expect.

holantomas:
@mabar I'm OK with that :+1:

src/Application/UI/Control.php
mabar: As I said, I understand. But imagine you don't know that code, just api. It looks exactly like a shortcut, but acts absolutely diferent.

holantomas: I don't, I'm not saying you're not right, but I think user don't care about it, he just want to set template file/view and this did that without unnecessary loads.

And I really hate setting template in `$onAnchor` event. Or do you know way how to do it without event in control factory or in `createComponent<name>()` method?

mabar: I think you could configure template file through it's factory definition
```
fooControlFactory:
  implement IFooControlFactory
  setup:
    - getTemplate()->setFile('/path/to/file')
```

But I never used that approach. I have ITemplate with setView method and to change path to these views I only need change pattern in configuration.

mabar:
I think you cannot as it requires more than ITemplate interface

src/Application/UI/Control.php
holantomas: This will end with `InvalidStateException` or new template instance. Anyway I don't like to setup file template in config, there are situations where you want to setup view by condition, or you have one control with factory and want to setup many different view for actual page.

milo:
IMHO, template file is internal thing of Control, thing of encapsulation. Control may use more templates too, that's why you usualy set template file in `render()` method.

Why do you need set template file outside of Control?

holantomas:
> IMHO, template file is internal thing of Control, thing of encapsulation. Control may use more templates too, that's why you usualy set template file in `render()` method.
> 
> Why do you need set template file outside of Control?

It's shown in forum topic. User usign 3rd party control like datagrid or paginator but HTML not equal to his UI for example. 

**encapsulation:**
`Control::getTemplate(): ITemplate` is public
`ITemplate::setFile()` is public
This doesn't break anything.

PR #212, opened 2019-03-14
Fix phpDoc of Component::redirect

janlanger:
- bug fix / new feature?   bugfix
- BC break? no

Parameter $code is deprecated and optional, but still present. The phpDoc should reflect what arguments really accepts.


mabar:
Can be closed, #223 will be used.

PR #219, opened 2019-05-15
Presenter: added onRender event

bckp:
Added new onRender event, right after beforeRender and before actual render happends (as put it before beforeRender and called it onRender is confusing)

- Type: new feature
- BC break: noe
- doc PR: none

If we want to use traits in project, and mainly in extensios, there should be easy way, how to push somethink into template, but only if template is used. We have onStartup and onShutdown methods, but they are not ideal for this, as onStartup is called even if no reder<action> is performed, so if i add somethink to template there, it will cost me time of template creation, that will never use if redirect is called.

We have beforeRender and afterRender, but that is only methods, so we can modify them with trait.

I suggest new onRender method, that is called right after beforeRender, before render happends (if we called it onBeforeRender, we can put it before beforeRender method, but for i belive onRender is self describe so everybody will understand that.

### DIFF
--- a/src/Application/UI/Presenter.php
+++ b/src/Application/UI/Presenter.php
@@ -51,6 +51,9 @@ abstract class Presenter extends Control implements Application\IPresenter
 	/** @var callable[]&(callable(Presenter $sender): void)[]; Occurs when the presenter is starting */
 	public $onStartup;
 
+	/** @var callable[]&(callable(Presenter $sender): void)[]; Occurs when the presenter is rendering after beforeRender */
+	public $onRender;
+
 	/** @var callable[]&(callable(Presenter $sender, IResponse $response): void)[]; Occurs when the presenter is shutting down */
 	public $onShutdown;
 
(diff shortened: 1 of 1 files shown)


src/Application/UI/Presenter.php
```diff
@@ -213,6 +216,7 @@ public function run(Application\Request $request): Application\IResponse
 
 			// RENDERING VIEW
 			$this->beforeRender();
+			$this->onRender($this);
```
mabar: Method is missing IResponse declared in phpdoc.
View and render method parameters could be useful.
If onRender, why not onAction?

bckp: PHPDoc: my bad, i writed it directly on GH.. will fix it...
I want onBeforeRender and onAfterRender, but after i spoke with DG, onRender is better, as onAfterRender have almost no usage...

bckp:
Removed IResponse from onRender (no reason why it should be there) and fixed CS

dg:
Thanks

PR #224, opened 2019-07-18
Fix $this to self:: in case of static method.

janbarasek:
- bug fix
- BC break? no

Simply safe replace `$this` to `self::` in case of static method.


dg:
It seems like BC break, it should probably be `static::`

PR #236, opened 2019-10-21
BadRequestException parent Exception expect Throwable

milous:
- bug fix
- BC break? no

https://www.php.net/manual/en/class.exception.php
```
public __construct ([ string $message = "" [, int $code = 0 [, Throwable $previous = NULL ]]] )
```

The third parameter can be only \Throwable


dg:
Thanks

PR #237, opened 2019-11-18
Create directory for touch file in case directory not exists

ricco24:
- bug fix
- BC break? no

### DIFF
--- a/src/Bridges/ApplicationDI/PresenterFactoryCallback.php
+++ b/src/Bridges/ApplicationDI/PresenterFactoryCallback.php
@@ -44,6 +44,10 @@ final class PresenterFactoryCallback
 
 		} elseif (!$services) {
 			if ($this->touchToRefresh) {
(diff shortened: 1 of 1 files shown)


src/Bridges/ApplicationDI/PresenterFactoryCallback.php
```diff
@@ -44,6 +44,10 @@ public function __invoke(string $class): Nette\Application\IPresenter
 
 		} elseif (!$services) {
 			if ($this->touchToRefresh) {
+				$dir = pathinfo($this->touchToRefresh, PATHINFO_DIRNAME);
```
dg: Maybe Nette\Utils\FileSystem::createDir(dirname($this->touchToRefresh)) is be better

ricco24: ok, changed

dg:
thanks

PR #238, opened 2019-11-26
FileResponse: added some checks

tacoberu:
- bug fix? yes
- BC break? no

If the file has strict privileges, it will crash the entire application.

### DIFF
--- a/src/Application/Responses/FileResponse.php
+++ b/src/Application/Responses/FileResponse.php
@@ -40,6 +40,9 @@ final class FileResponse implements Nette\Application\IResponse
 		if (!is_file($file)) {
 			throw new Nette\Application\BadRequestException("File '$file' doesn't exist.");
 		}
+		if (!is_readable($file)) {
+			throw new Nette\Application\BadRequestException("File '$file' is not readable.");
+		}
 
 		$this->file = $file;
 		$this->name = $name ?? basename($file);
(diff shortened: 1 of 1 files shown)


src/Application/Responses/FileResponse.php
```diff
@@ -40,6 +40,9 @@ public function __construct(string $file, string $name = null, string $contentTy
 		if (!is_file($file)) {
 			throw new Nette\Application\BadRequestException("File '$file' doesn't exist.");
 		}
+		if (!is_readable($file)) {
+			throw new Nette\Application\BadRequestException("File '$file' doesn't readable.");
```
dg: is not readable sounds better

src/Application/Responses/FileResponse.php
```diff
@@ -88,6 +91,9 @@ public function send(Nette\Http\IRequest $httpRequest, Nette\Http\IResponse $htt
 
 		$filesize = $length = filesize($this->file);
 		$handle = fopen($this->file, 'r');
+		if (!$handle) {
+			throw new Nette\Application\BadRequestException("Cannot make handle of '{$this->file}'.");
+		}
```
dg: Cannot open file?

src/Application/Responses/FileResponse.php
JanTvrdik: I'm not sure how useful the `is_readable` check is when you check `fopen` result later anyway.

tacoberu: @JanTvrdik The sooner the better :-) But I don't insist on it.

PR #239, opened 2019-12-02
Presenter: injectPrimary() params are required because $httpRequest is required

janbarasek:
- new feature?
- BC break? no

Mark optional parameters as required (without `= null`), because parameter `$httpRequest` is required.

Removed set default value as `null`, because in real code user must define value. Only marked as nullable.


dg:
Thanks

PR #243, opened 2020-01-02
Presenter: added isModuleCurrent() [Closes #240]

### LINKED ISSUE #240: introduce isModuleCurrent method
**Is your feature request related to a problem?**
- I can use $presenter->isLinkCurrent('MyModule:MyPresenter:*')
- using $presenter->isLinkCurrent('MyModule:*') throws exception

I guess that many had tried to use the analogy for modules - but $presenter->isLinkCurrent('MyModule:*') throws InvalidLinkException. That looks correct to me after thinking for a while - trying to link module (not a presenter from the module) is meaningless.

Anyway I believe there should be a way how to determine the same for modules - can be handy in navigation etc.

**Explain your intentions.**
Provide a default solution for simple determination that I'm linking from current application module.

**It's up to you to make a strong case to convince the project's developers of the merits of this feature.**
When framework encourages use of modules, it can be confusing (newcomers included) that one can link to a submodule, can determine linking to current presenter but get an error when using similar notation for modules.

When looking for solution, I found two ([first](https://forum.nette.org/cs/15963-presenter-islinkcurrent-na-modul), [second](https://forum.nette.org/cs/11774-zvyrazneni-aktualniho-modulu-v-menu)) really old related posts in forum.

Anyone can come up with his own solution so when this request is considered inappropriate, I would at least make a remark in documentation (maybe [here](https://doc.nette.org/en/3.0/presenters#toc-link-creation)) that it cannot be used in this way.

khorsky:
- new feature
@dg has already added `$presenter->getModule()` method, which is universal. I've added an alternative to `$presenter->isLinkCurrent()` for modules.

- doc PR: nette/docs#???  I'd send PR if accepted. I've checked current doc and send small PR (https://github.com/nette/docs/pull/790). I would propose to describe new behavior in same section.

### DIFF
--- a/src/Application/UI/Presenter.php
+++ b/src/Application/UI/Presenter.php
@@ -188,6 +188,18 @@ abstract class Presenter extends Control implements Application\IPresenter
 	}
 
 
+	public function isModuleCurrent(string $moduleName): bool
+	{
+		if (!$moduleName) {
+			throw new Nette\InvalidArgumentException('Missing module name!');
+		}
+
+		$moduleName = trim($moduleName, ':') . ':';
+
+		return Nette\Utils\Strings::startsWith($this->getModule() . ':', $moduleName);
+	}
+
+
 	/********************* interface IPresenter ****************d*g**/
 
 
--- a/tests/UI/Presenter.isModuleCurrent().phpt
+++ b/tests/UI/Presenter.isModuleCurrent().phpt
@@ -1,0 +1,67 @@
+<?php
+
+/**
+ * Test: Presenter::isModuleCurrent.
+ */
+
+declare(strict_types = 1);
+
+use Tester\Assert;
+
+require __DIR__ . '/../bootstrap.php';
+
+
+class TestPresenter extends \Nette\Application\UI\Presenter
+{
+}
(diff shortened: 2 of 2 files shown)


dg:
Thanks.

What do you think about change name to `isInModule()` (or something similar), because `isModuleCurrent` sounds to me like an exact match.

khorsky:
In template I would read `$presenter->isModuleCurrent(<some-module-name>)` as "is given module current" - that seems logical to me. 

Your proposal makes sense too (line of code saying "is this presenter in the module XY"). Despite that I do not "like it". When trying to find a better name I refused
- isInModule - does not work for me, see bellow
- isSubmodule - tells another story
- isModulePart / isModuleRelated - these are desperate attempts

I would stick to `$presenter->isModuleCurrent`. It came to my mind as first option when I was looking for the behavior (I was looking for "isLinkCurrent for modules").

It's similar in name and similar in usage. The most beneficial use case is probably to mark navigation item as active/current. I tried to use "isInModule" in the template with this use case - it looks strange...

dg:
Ok :-)

PR #245, opened 2020-01-09
Control, Presenter::createTemplate() added parameter $class for custom creation of the Template

solcik:
- new feature
- BC break? no
- doc PR: nette/docs#???  

Based on the discussion in https://forum.nette.org/en/32893-custom-template-class-for-presenter-and-control-experimental-feature#p207624, I added parameters to these methods. Is it ok like this?


dg:
Great, thanks

PR #247, opened 2020-01-25
ComponentReflection: saveState - unset instead of setting null

peldax:
- bug fix  
- BC break?    maybe, needs review

Save state methods removes parameter entirely when parameter value matches default value.

In current version, parameter value is set to null, browser than receives empty value, which is not intended.


peldax:
For example: My datagrid has persistent $page with default value 1. SaveState nulls the value. My javascript remembers state recieved by nette and sends it back on next request, which results on error.

PR #256, opened 2020-03-24
LinkGenerator: allow semicolon destination prefix

### LINKED ISSUE #255: Inconsistent reaction to deprecated link's destination format
Version: 3.0.4

Links with destination in deprecated/invalid format\
`:Module:Presenter:action` (note the semicolon `:` at the beginning of the destination string)\
work with `UI\Presenter`/`UI\Component` but not with `LinkGenerator`. 

I'm not sure which one is the desired behaviour, or whether this is a bug of `UI\Component` or `LinkGenerator` class, but in the following scenario, only the first call works:
```php
$presenter->link(':Foo:Bar:'); // works ✔
$linkGnrtr->link(':Foo:Bar:'); // fails ❌
```
The second call fails with `Presenter name must be alphanumeric string, ':Foo:Bar' is invalid.`.

Links in the correct format work with both classes, of course ✔:
```php
$presenter->link('Foo:Bar:'); // works ✔
$linkGnrtr->link('Foo:Bar:'); // works ✔
```

I understand that the format is not correct according to current documentation, but I would like to point out that the behaviour should be _consistent_, that is,\
**it should either work in both cases or not work in either one**.

dakujem:
- fixes bug #255 
- BC break? no

Allows passing semicolon-prefixed destinations to `LinkGenerator::link` calls:
```php
(new LinkGenerator(...))->link(':Foo:Bar:baz');
(new LinkGenerator(...))->link(':Bar:baz');

// and also this:
(new LinkGenerator(...))->link($presenter->getAction(true));
```

--
closes #255


dakujem:
@dg I'd be grateful if this was merged & released promptly. It would be awesome. I see praparations for next release. This should be fully compatible.

dakujem:
Should I rebase this? Any plan to fix the issue?

dakujem:
You should update your test config - PHP 7.1 is no longer your target.

dakujem:
Do we really need to force-push to `master`? Isn't that what dev branches are meant for? I've needed to rebased this PR 4 times since I originally prepared it. It would be awesome to accept it already.

PR #259, opened 2020-04-21
RoutingPanel: Key should be together.

janbarasek:
- bug fix
- BC break? no

Fix cases like this:

Snímek obrazovky 2020-04-21 v 12 21 35

Thanks.


dg:
It looks like the current version is no longer affected. If so, reopen it.

PR #260, opened 2020-04-24
RoutingPanel: Add support for render routes with exception.

janbarasek:
- new feature
- BC break? no

I think in case of some Custom router throws exception it can be rendered in Tracy panel.

Current behavior (with callstack):

Snímek obrazovky 2020-04-24 v 10 01 40

New behavior (with marked route and `x` symbol):

Snímek obrazovky 2020-04-24 v 10 22 29

I think this is more better. :)

Thanks.


dg:
Thanks

PR #262, opened 2020-05-28
Make Presenter::createRequest non-final

dakujem:
- BC break? **no**

**Summary**\
Make `Presenter::createRequest()` method non-final.
Furthermore, inside `Presenter::createRequest()` call `static::argsToParams()` instead of `self::argsToParams()` to allow for extension.

**Rationale**\
The rationale behind the change is simple: **allow for extension** of link creation.\
For well over 6 years, since `2.0.14`, it has been non-final, and now it is final in `3.0` again. It was final in `2.0.*`, then made non-final in `2.0.14` just half a year later. It was final in `0.*` versions, but those were deep medieval ages.

**Use cases**\
I now have two use cases that require the method's extension.
- _custom link logic_ for which i found a workaround (not an ellegant one)
- and now I am unable to implement _method injection_ because of links requiring all `action*` method's arguments being passed.

Thank you for considering this.


dakujem:
bump

dakujem:
Finally, thank you.

PR #225, opened 2019-07-30
RoutingPanel: Add better panel responsivity

janbarasek:
- new feature
- BC break? no

My suggestion for better responsivity in case of resize window. This feature is very helpful when you must resize Tracy panel in small notebook screen or on mobile phone in case of debug by frontend developer in company local network.

New features:

1. Main table with `width: 100 %` for better user experience.

Snímek obrazovky 2019-07-30 v 13 59 56

2. In case of medium size wrap lines around slashes in `Mask / Class` by `<wbr>` and keep slash before string on each new lines.

Snímek obrazovky 2019-07-30 v 14 00 49

3. Wrap URL parameters or path or both:

Snímek obrazovky 2019-07-30 v 14 02 10

Snímek obrazovky 2019-07-30 v 14 02 20

Snímek obrazovky 2019-07-30 v 14 02 33

Thanks.

PR #265, opened 2020-08-14
Better look of RoutingPanel

petrparolek:
- bug fix
- BC break? no

before
obrazek

after
obrazek


petrparolek:
@dg what do you think about it?

dg:
`table { width: 100% }` should be default.

The other changes go against #225.

PR #267, opened 2020-11-09
Prevent TypeError

ondrejmirtes:
- bug fix 
- BC break? no

If someone constructs a URL that looks like `example.com/?1-foo=bar`, Nette crashes with this error:

> Argument 1 passed to Nette\ComponentModel\Container::getComponent() must be of the type string, int given, called in vendor/nette/application/src/Application/UI/Presenter.php on line 217


ondrejmirtes:
Build failure unrelated.

dg:
Thanks

Issue #271, opened 2021-01-07
flashMessage() accepting general interface


dakur:
- Is your feature request related to a problem? Please describe.

Rendering `HtmlStringable` in flash message.

- Explain your intentions.

We have class `SanitizedHtml` which `implements HtmlStringable` and which we use to render safe HTML into template. Often, we need to render HTML in flash message as well (e.g. bold text). That's where we call `$this->flashMessage($sanitizedHtmlInstance)`

But phpstan complains `Parameter #1 $message of method Nette\Application\UI\Control::flashMessage() expects stdClass|string, Nette\HtmlStringable given.`

I suggest making an interface for flash message parameter objects so that the type would be `\stdClass|string|FlashMessage` which our `SanitizedHtml` could implement then.

- It's up to you to make a strong case to convince the project's developers of the merits of this feature.
😅

dakur:
Or maybe just add `HtmlStringable` instead of creating `FlashMessage` interface..

PR #272, opened 2021-01-07
flashMessage: allow HtmlStringable as a message

jkuchar:
- bug fix for #271
- BC break? no (fixes BC break)

This allows behaviour which was supported before phpdoc was added. (there used to be `mixed` type for this parameter)

cc @dakur


jkuchar:
Broken build errors are not related to this MR. It looks like some unrelated changes from master.

PR #275, opened 2021-01-26
LatteExtension: fix undefined property

mabar:
- bug fix
- BC break? no

Current version fails in case of `!isset($template->control) && $all === true`.
This PR fixes the problem and adds support for `$template->presenter` in case that `$template->control` is undefined


mabar:
Build failures seems unrelated to the PR

dg:
Thanks

PR #277, opened 2021-02-04
Presenter: try to find Template file for each action

lulco:
- new feature
- BC break? no
- doc PR: not yet

For presenters with multi actions we should create Template for each of them. Based on https://forum.nette.org/en/32893-custom-template-class-for-presenter-and-control-experimental-feature#p207580


dg:
Good idea

PR #283, opened 2021-02-16
LatteExtension: changed UI\TemplateFactory type to ApplicationLatte\TemplateFactory

mabar:
Latte TemplateFactory have `createTemplate()` extended for `$class` parameter and more accurate return typehint, which are both frequently used. This allows `TemplateFactory` to be autowired by `ApplicationLatte\TemplateFactory` type and prevent users confusion (like [here](https://forum.nette.org/en/34214-how-to-use-callbacks-in-services) and [here](https://forum.nette.org/cs/34209-factory-problem-s-construct))

Also some hint explaining why autowiring does not work when user sees in its DIC that service of requested type is available could be super helpful, but I have no idea how complicated the implementation would be

### DIFF
--- a/src/Bridges/ApplicationDI/LatteExtension.php
+++ b/src/Bridges/ApplicationDI/LatteExtension.php
@@ -70,7 +70,7 @@ final class LatteExtension extends Nette\DI\CompilerExtension
 		}
 
 		$builder->addDefinition($this->prefix('templateFactory'))
-			->setType(Nette\Application\UI\TemplateFactory::class)
+			->setType(ApplicationLatte\TemplateFactory::class)
 			->setFactory(ApplicationLatte\TemplateFactory::class)
 			->setArguments(['templateClass' => $config->templateClass]);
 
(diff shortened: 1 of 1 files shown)


dg:
So I hope it's not a BC break.

src/Bridges/ApplicationDI/LatteExtension.php
```diff
@@ -70,7 +70,7 @@ public function loadConfiguration()
 		}
 
 		$builder->addDefinition($this->prefix('templateFactory'))
-			->setType(Nette\Application\UI\TemplateFactory::class)
+			->setType(ApplicationLatte\TemplateFactory::class)
```
dg, line 73: This line can be removed

mabar:
Afaik shouldn't be? Unless user has a factory set to something what cannot be used directly as a type, it should behave in BC compatible manner.

dg:
https://xkcd.com/1172/

PR #288, opened 2021-05-21
LatteExtension: inject interface to be more compatible

### LINKED ISSUE #287: Unable to use custom TemplateFactory with LatteExtension since 3.1
Version: `3.1.*`

Since 3.1, it is impossible to use a custom implementation of `Nette\Application\UI\ITemplateFactory` (`Nette\Application\UI\TemplateFactory` since 3.1) with `Nette\Bridges\ApplicationDI\LatteExtension`.

### Steps To Reproduce
1. implement `ITemplateFactory`
2. register it as your template factory service (`latte.templateFactory`)
3. try to compile the container

### What's going on

[LatteExtension registers an on-compile call to `LatteExtension::initLattePanel`](https://github.com/nette/application/blob/v3.1/src/Bridges/ApplicationDI/LatteExtension.php#L97) that has a type hint for the concrete implementation `Nette\Bridges\ApplicationLatte\TemplateFactory` and not the interface, causing `TypeError`.

milo:
- bug fix #287
- BC break? no

I hit the #287 too. After Latte upgrade and with my own `UI\TemplateFactory` implementation in DI (with `ApplicationLatte\TemplateFactory` autowiring off) container buid fails.

I thik this fix will work but it's a pitty that one lost new great latte panel. So it is more or less hotfix. Maybe to find another way how to hook the panel to template, maybe via decorator, but I'm sure how to do it.

### DIFF
--- a/src/Bridges/ApplicationDI/LatteExtension.php
+++ b/src/Bridges/ApplicationDI/LatteExtension.php
@@ -94,17 +94,19 @@ final class LatteExtension extends Nette\DI\CompilerExtension
 	}
 
 
-	public static function initLattePanel(ApplicationLatte\TemplateFactory $factory, Tracy\Bar $bar, bool $all = false)
+	public static function initLattePanel(Nette\Application\UI\TemplateFactory $factory, Tracy\Bar $bar, bool $all = false)
 	{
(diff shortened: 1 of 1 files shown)


dg:
Thanks

dakujem:
@dg Could you please merge this into the `v3.1` branch too?

Issue #290, opened 2021-07-02
Methods like sendResponse() should contain never-return typehint


janbarasek:
Hello,

I think methods like `sendJson()` should contain PHPStan annotation `@phpstan-return never-return` like this:

```php
/**
 * Sends JSON data to the output.
 * @throws Nette\Application\AbortException
 * @phpstan-return never-return
 */
public function sendJson(mixed $data): void
{
	$this->sendResponse(new Responses\JsonResponse($data));
}
```

Currently, this can generate hundreds of errors when PhpStan reports possible undefined variables. The annotation marks the code after this call as dead.

For example:

```php
public function actionDefault(): void
{
    if ($this->variable) {
        $a = 1;
    } else {
        $this->sendJson('error');
    }

    echo $a; // generate possible undefined variable, but it is not true
}
```

Yes, there is an official extension for PhpStan and Nette framework that does this solution, but it would be good to have native support as well.

Thanks!

JanTvrdik:
The official type name is just `never`.
https://github.com/php/php-src/pull/6761

dg:
`@return never` seems good, can you send PR?

janbarasek:
Solved by #291

PR #291, opened 2021-07-07
Presenter: Add return never typehint. (#291) (#290)

janbarasek:
- new feature? #290
- BC break? no

Just add `return never` typehint for PhpStan.

### DIFF
--- a/src/Application/UI/Presenter.php
+++ b/src/Application/UI/Presenter.php
@@ -1044,7 +1044,6 @@ abstract class Presenter extends Control implements Application\IPresenter
 
 	/**
 	 * Restores request from session.
-	 * @return never
 	 */
 	public function restoreRequest(string $key): void
 	{
--- a/src/Application/UI/Presenter.php
+++ b/src/Application/UI/Presenter.php
@@ -661,7 +661,6 @@ abstract class Presenter extends Control implements Application\IPresenter
 	/**
 	 * Conditional redirect to canonicalized URI.
 	 * @throws Nette\Application\AbortException
-	 * @return never|void
 	 */
 	public function canonicalize(string $destination = null, array $args = []): void
 	{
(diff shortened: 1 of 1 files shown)


src/Application/UI/Presenter.php
```diff
@@ -654,6 +661,7 @@ final public function getLastCreatedRequestFlag(string $flag): bool
 	/**
 	 * Conditional redirect to canonicalized URI.
 	 * @throws Nette\Application\AbortException
+	 * @return never|void
```
dg, line 664: @JanTvrdik Can be never combined?

mabar: It can't be https://github.com/php/php-src/blob/master/Zend/tests/type_declarations/union_types/never_with_class.phpt

janbarasek: Okay, but how do we solve this? The method often returns `never`, but sometimes `null`. Neither typehint is correct.

mabar: > In type theory noreturn would be called a “bottom” type. That means it's effectively a subtype of every other type in PHP’s type system, including void.
https://wiki.php.net/rfc/noreturn_type#variance

Void is correct.

janbarasek: Ok, reverted back to `void`.

src/Application/UI/Presenter.php
```diff
@@ -1037,6 +1044,7 @@ public function storeRequest(string $expiration = '+ 10 minutes'): string
 
 	/**
 	 * Restores request from session.
+	 * @return never
```
dg, line 1047: Here should be void too.

janbarasek: Fixed.

dg:
Great, thanks

PR #296, opened 2021-09-25
LinkGenerator: accept classname in $dest

mabar:
- new feature
- BC break? no

This is minimal change needed in Nette in order to support NOT using mapping and to reference class directly instead. 

```php
$this->linkGenerator->link(self::class . ':default');
```

To fully support that behavior also a `class_exists()` check is needed in `PresenterFactory`. Without this is my change untestable, so let me know if I should modify also `PresenterFactory` and write tests.
https://github.com/orisai/nette-application/blob/73906f0f3009f751c15d8170467246957ca2e239/src/Mapping/DefaultPresenterFactory.php#L147-L163

I am already able to achieve it with `$this->presenter->link($this->presenter::class . ':default');`.

Benefit of this change is presenters don't have to follow any namespace structure, just like any class.


dg:
There should be a condition that always makes it clear whether it is a class or a presenter, otherwise it could get awkward. For example, a class must contain at least one character `\`.

mabar:
I wrote an regex which expects `[[Class\Name:]action] [#fragment]` and all the tests on regex101.com work (check the unit tests tab). But I cannot get it work in PHP, it does not match my test string `App\Admin\User\List\UserListPresenter:default`, not sure why :/
https://regex101.com/r/u2DneV/1

mabar:
Okay, got it. Had to add some character classes. Seems like regex101 does not use php, just PCRE2 - regex with `[\\]` and `\\` works with regex101, `[\\\]` works with php

PR #304, opened 2022-06-10
Add `@return never` to redirect methods

spaze:
- bug fix / new feature?   no / not really
- BC break? no

Adds `@return never` to `Component::redirect()` and `Component::redirectPermanent()`, useful for static analysis and required when any method using either of those wants to use `never` as its return type declaration (`function foo(): never`).

`Presenter::redirectUrl()` already [uses it](https://github.com/nette/application/blob/25e61b0de03bdb75a79ce5c72badcd4d3d6c4f01/src/Application/UI/Presenter.php#L609), this PR "propagates" it further.

Thank you!

### DIFF
--- a/src/Application/UI/Component.php
+++ b/src/Application/UI/Component.php
@@ -304,6 +304,7 @@ abstract class Component extends Nette\ComponentModel\Container implements Signa
 	 * Redirect to another presenter, action or signal.
 	 * @param  string   $destination in format "[//] [[[module:]presenter:]action | signal! | this] [#fragment]"
 	 * @param  mixed  ...$args
+	 * @return never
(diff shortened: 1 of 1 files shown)


spaze:
Hi @dg, do you have a plan when this will be released? Thanks!

dg:
It is in v3.1.4

spaze:
3.1.4 was released before this PR was authored, maybe you mean #291 but that's a different PR. This one wasn't released yet.

dg:
You're right, I guess I lost this commit. I'll release it in the next patch version.

spaze:
Happens :-) Thank you!

Issue #307, opened 2022-08-31
Ugly look of Tracy routing bar


petrparolek:
Version: 3.1-dev

### Bug Description
Tracy routing bar has ugly look

### Steps To Reproduce
![obrazek](https://user-images.githubusercontent.com/6066243/187643510-c1416b2a-071d-461e-a2a9-b81e1d65d1e8.png)

### Expected Behavior
![obrazek](https://user-images.githubusercontent.com/6066243/187643692-18feb909-d7af-4c85-b750-5d79ef16720b.png)

### Possible Solution
```diff --git a/src/Bridges/ApplicationTracy/templates/RoutingPanel.panel.phtml b/src/Bridges/ApplicationTracy/templates/RoutingPanel.panel.phtml
index 61839780..c9808a50 100644
--- a/src/Bridges/ApplicationTracy/templates/RoutingPanel.panel.phtml
+++ b/src/Bridges/ApplicationTracy/templates/RoutingPanel.panel.phtml
@@ -45,6 +45,7 @@ use Tracy\Helpers;
        #tracy-debug .nette-RoutingPanel pre, #tracy-debug .nette-RoutingPanel code {
                display: inline;
                background: transparent;
+               white-space: nowrap;
        }
 
        #tracy-debug .nette-RoutingPanel-rel {
```

PR #308, opened 2022-08-31
Polished look of routing Tracy panel

petrparolek:
- bug fix (issue https://github.com/nette/application/issues/307)
- BC break? no

Before:
obrazek

after:
obrazek


dg:
I see it fine on my end. Somehow you're getting the style `word-break: break-all` in the right column, can you find out where?

PR #309, opened 2022-10-18
Presenter::$onAfterSignal event

mabar:
- new feature
- BC break? no

I am currently working on a Tracy bar which tracks submitted forms and shows errors from these forms.

This is useful for debugging manually rendered forms which don't render all form errors. e.g. datagrids

Problem is it is currently hooked via `onStartup` event. That works only for components that are created via `createComponent*()`, have all required parameters for creation available during startup and are not created manually - e.g. via `$this['form'] = new Form();` in `action*()` method. It also changes order of code execution.

Having an `onAfterSignal` event would solve all of these issues and would also allow me to handle successfully submitted forms (which throw AbortException for redirect) and other signals

Current (bad) solution:
- https://github.com/orisai/nette-application/blob/bc394c43189e84f26f2f6efef2ff5de71414b7c5/src/FormMonitor/DI/FormMonitorExtension.php#L116-L122
- https://github.com/orisai/nette-application/blob/bc394c43189e84f26f2f6efef2ff5de71414b7c5/src/FormMonitor/SignalFormExtractor.php#L22

If you agree on proposed solution @dg I will write some tests for it.

### DIFF
--- a/src/Application/UI/Presenter.php
+++ b/src/Application/UI/Presenter.php
@@ -15,6 +15,7 @@ use Nette\Application\Helpers;
 use Nette\Application\Responses;
 use Nette\Http;
 use Nette\Utils\Arrays;
+use Throwable;
 
 
 /**
@@ -51,6 +52,9 @@ abstract class Presenter extends Control implements Application\IPresenter
 
 	/** @var array<callable(self): void>  Occurs when the presenter is starting */
 	public $onStartup = [];
+
+	/** @var array<callable(self, Throwable|null): void>  Occurs after presenter processed a signal */
+	public $onAfterSignal = [];
 
 	/** @var array<callable(self): void>  Occurs when the presenter is rendering after beforeRender */
 	public $onRender = [];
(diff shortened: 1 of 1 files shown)


src/Application/UI/Presenter.php
```diff
@@ -230,7 +234,17 @@ public function run(Application\Request $request): Application\Response
 
 			// SIGNAL HANDLING
 			// calls $this->handle<Signal>()
-			$this->processSignal();
+			$signalException = null;
+			try {
+				$this->processSignal();
+			} catch (Throwable $signalException) {
```
dg, line 240: I think the catch can be omitted

dg:
Seem legit.

For consistency, it might be a good idea to add $onSignal too.

src/Application/UI/Presenter.php
mabar: It is used to assign Throwable to $signalException

dg: Remove it too. Just try ... finally

PR #312, opened 2023-02-08
not correct path to TextResponse class

SergioPylyp:
I get massage error that Class "Nette\Application\UI\Responses\TextResponse" not found.  only writing correct path to this class help to fix it

- bug fix / new feature?   bug fix
- BC break? yes/no
- doc PR: nette/docs#???


dg:
There is `use Nette\Application\Responses;`

PR #317, opened 2023-10-05
exception in application: pass reference to previous presenter

dakur:
- new feature
- BC break? no
- doc PR: n/a

I need separate handling of bad request of presenters that implement some interface. To check this in error presenter, I need instance of the original presenter therefore this PR. It can be probably achieved also via reflection, but this it seems cleaner to me.

todo:
- [x] update tests


mabar:
So, you don't need an instance? 

```php
class ErrorPresenter extends Presenter 
{
	public function __construct(private readonly IPresenterFactory $presenterFactory) {}

	public function actionDefault(Throwable $exception, Request|null $request): void
	{
		$presenterName = $request?->getPresenterName() ?? null;
		if ($presenterName !== null) {
			$presenterClass = $this->presenterFactory->getPresenterClass($presenterName);
			dump(is_a($presenterClass, CheckedInterface::class, true));
        }
	}

}
```

dg:
Isn't it easier to inject an Application into the errorpresenter?

dakur:
> So, you don't need an instance?

No. Even though your solution work, it seems strange to me to work with presenter *factory* inside of presenter. Extracting the presenter router<->class name translation into separate single responsibility service would probably solve that. Does it make sense?

> Isn't it easier to inject an Application into the errorpresenter?
What would I do then? I can see `getRequests()` but still you have to call `$presenterFactory->getPresenterClass()`.

My proposal seems quite straightforward to me, is there any problem with it?

dakur:
Thank you!

PR #328, opened 2024-04-10
Extend syntax used in link() method of LinkGenerator

mildabre:
Proposed changes make possible to use in LinkGenerator the same syntax as in presenter - links started with //: or :

In the current version of LinkGenerator it is not possible to use for $destination in method link() syntax :Module:Presenter:action or  //:Module:Presenter:action as in presenter.

The proposed changes make possible to use this syntax in LinkGenerator also, withou any BC break, because the current syntax Module:Presenter:action will remain in function.

The proposal is also discussed on Nette formu in this post: https://forum.nette.org/cs/36468-rfc-rozsirit-syntaxi-parametru-destination-v-metode-linkgenerator-link#p226631


mabar:
Also discused in #255
This would be really welcomed feature :)

Some tests would be nice - https://github.com/nette/application/blob/30d90f4fca1df8face66b6963308935319413bfd/tests/Routers/LinkGenerator.phpt

Btw, whats the `@` for? Never saw that syntax

stpnkcrk:
@mabar experimental link aliases

https://forum.nette.org/en/36451-nette-application-3-2-aliases-for-links

dg:
In branch 3.2-dev the link generation in presenters and LinkGenerator is merged and the syntax is identical.

PR #329, opened 2024-05-04
Missing link from RoutingPanel.php to presenter class file in IDE

mildabre:
RoutingPanel displays name of current presenter with current action method and link to presenter class file in IDE. In case that method of current presenter doesnt exists - this link is missing. Proposed change of code fixes this small bug.

- bug fix YES
- BC break? NO
- doc detailed description on nette forum: https://forum.nette.org/cs/36505-poopravit-proklik-z-routingpanelu-do-souboru-presenteru-v-ide#p226916


dg:
Thanks

PR #332, opened 2024-05-28
Correction of false exception message in startup check

mildabre:
The exception message states the lowest possible presenter class where method beforeRender is found, so the error can be here or in its parents not descendants !!

- bug fix
- BC break - no
- doc PR: ---


dg:
It should probably say "ancestor"

mildabre:
Yes, I agree, "ancestor" is good solution

PR #338, opened 2024-11-17
Cast "true"/"false" to bool

thorewi:
- new feature
- BC break: no
- doc PR: probably not needed

For API routes/actions, It's really useful to cast "true" / "false" values to boolean and thus allow them to use for boolean parameters, because many other languages cast boolean to true/false and not 0/1 and then you are getting these "true"/"false" arguments.

### DIFF
--- a/src/Application/UI/ParameterConverter.php
+++ b/src/Application/UI/ParameterConverter.php
@@ -160,6 +160,16 @@ final class ParameterConverter
 			return false;
 		}
 
+		if ($type === 'bool') {
+			if ($val === 'true') {
+				$val = true;
+				return true;
+			} elseif ($val === 'false') {
+				$val = false;
+				return true;
+			}
+		}
+
 		$tmp = ($val === false ? '0' : (string) $val);
 		if ($type === 'float') {
 			$tmp = preg_replace('#\.0*$#D', '', $tmp);


mabar:
Commonly handled strings for this usecase are also True, False, 1 and 0

thorewi:
"1" and "0" is already correctly handled...

if we want to support also "True"/"False" or "yes"/"no" and similar, we can maybe use filter_var, like this:

filter_var($string, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE)

some examples:

https://onlinephp.io?s=s7EvyCjg5eLlSk3OyFcoSyyKTynNLdBIy8wpSS2KB_I1lEqKSlOVdBTcPH1CXIPiwxx9PF0cQ1zjnfz9fVwd_eASfqE-PvH-fvFujp4-oUGumprW-AwNoYmhQaGuVDc0LTGnmPpOdaONqY4-wdQPgcrUYqqbGUkLM12DqW5mXj7VjfSjgZH-VDfSkOomGpBpIl5X6pFrKh530sJMA-qbaWhIbhwBAA%2C%2C&v=8.2.20

dg:
Such a change is not possible because it would be a colossal BC (backward compatibility) break, where old URLs would start to have a different meaning.

thorewi:
Teoretically yes, practically I don't know... 

Right now when someone proceeds to ?aaa=true he gets an error, after this change he will be autocanonicalize to ?aaa=1 and will get the same as he would proceeds directly to ?aaa=1...

no idea if it could be such an issue/BC in real world...

To avoid speculation, maybe we can make this behaviour switchable?

mabar:
Oh, this is nette/application, not schema 🙈
I think better for advanced mapping would be to map the request to a typed object in which validation rules may be configurable and therefore allow mapping of bool-like types, database entities etc.
Symfony quite recently added something similar. Now I am just thinking about how to make it work seamlessly with the component model in Nette

PR #347, opened 2025-06-14
Correcting Misleading Protected Modifiers in Presenter class

mildabre:
In the ApplicationExtension class, the checkPresenter() method enforces that the presenter methods startup(), beforeRender(), afterRender(), and shutdown() must be public and non-static. However, in the base abstract Presenter class within the Nette framework, these methods are defined as protected. While this does not prevent them from being overridden in derived classes, it introduces unnecessary confusion and lacks clear justification.

After upgrading my projects to the latest version of Nette, ApplicationExtension started throwing exceptions such as:

Nette\InvalidStateException: Method Nette\Application\UI\Presenter::checkHttpMethod(): this method must be public non-static.

This issue occurred only in some older projects that had not been updated for a long time, even though all final presenters had correctly set modifiers as public, while the majority of my updated projects remained unaffected. I could not determine why certain projects triggered the exception while others did not. However, after manually refactoring the Presenter class inside /vendor/ and changing the visibility of the mentioned methods from protected to public, the issue was completely resolved—without any loss of normal presenter functionality.

Proposed Change
I propose modifying the affected methods in the nette/application package by changing their visibility from protected to public, for two key reasons:

a) eliminating unnecessary confusion - the protected modifier in this context serves no meaningful purpose and only introduces ambiguity.

b) preventing unnecessary exceptions - under specific conditions, even syntactically correct final presenters can unexpectedly trigger errors, which this change would effectively prevent.

I opened this theme for public discussion on [forum nette](https://forum.nette.org/en/36863-correction-of-misleading-protected-modifiers-in-presenter-class-proposal)


mildabre:
By reviewing the incriminated project, I found that the protected modifiers were not the cause of the false exception being thrown. However, the first and main argument for the change remains, the simplification of unnecessarily complex code.

PR #355, opened 2025-08-20
LinkGenerator: do not pass defaults of #[Parameter]

vrana:
- new feature
- BC break? no
- doc PR: n/a

With `render($x = 1)`, no URL parameter is added when linking to `x => 1`. With `#[Parameter] public $x = 1`, it was added. This PR fixes it.


dg:
Thanks

